PR #78, opened 2024-05-10
feat: swap to qtpy and support pyside6

lucaseck:
We were locked to just Pyside2 support which limited the submitter to only being able to run with Python < 3.11.
Use qtpy instead so that Pyside2 can still be used, or Pyside6, or any other Qt package.
More flexibility and Pyside6 can now be used(`pip install deadline[gui]` brings in pyside6-essentials)
Ran the submitter using both Pyside2 and Pyside6 to submit a job
No
No
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,8 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 
 ### Using the KeyShot Submitter
 
-1. Install deadline-cloud and PySide2
+1. Install deadline and Pyside6
+    - e.g. `pip install deadline[gui]`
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -25,7 +25,8 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 
 ### Using the KeyShot Submitter
 
-1. Install deadline-cloud and PySide2
+1. Install deadline and Pyside6
+    - e.g. `pip install deadline deadline[gui]`
```
epmog:
```suggestion
    - e.g. `pip install "deadline[gui]"`
```

epmog (changes requested):

epmog (approved):

crowecawcaw (approved):

PR #87, opened 2024-06-03
rfc: replace submitter GUI with Deadline's bundle GUI submit

crowecawcaw:
In the current submitter design, the users runs a Python script in Keyshot when collects some info from the scene, then opens a second Python GUI application running in a separate Python installation. When the user clicks submit, the second application uses the Deadline library to actually submit the job.

The design is complex, and it requires the user to go through a tricky process to install the plugin and run Keyshot each time with the necessary environment variables.

Request for comments: have the Keyshot script call `deadline bundle gui-submit` directly with the info it collects and remove the whole second GUI application along with its dependencies and install process.

The new install process is:
1. Save this script to your Keyshot scripts folder.

Simpler code and install process. Loses a couple UX flourishes in the custom GUI application like previewing the output file path name or friendlier labels for output formats.

Submitting a job. Tests, docs, etc. haven't been updated. Looking for feedback on the initial shape of this first.

Not yet

Nope

----
Screenshot 2024-06-03 at 4 09 21 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ AWS Deadline Cloud for KeyShot is a python package that allows users to create [
 ## Compatibility
 
 This library requires:
-1. KeyShot 2023
+1. KeyShot 2023 or 2024
 1. Python 3.9 or higher; and
 1. Windows, or a macOS operating system.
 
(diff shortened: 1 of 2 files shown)


crowecawcaw:
Updated the code to use the adapter-less job template. The entire Keyshot integration now fits in a single file, and installation looks like saving the single script file to Keyshot's scripts folder. The adapter-less integration requires Keyshot to reopen the scene between every frame. Sharing as a POC.

Also included: user-edited settings are persisted between job submissions.

PR #93, opened 2024-06-18
feat!: replace submitter GUI with bundle GUI submit

crowecawcaw:
The submitter Python script currently calls a second Python application to show the submitter GUI. This setup requires a second Python installation, manually set environment variables, and extra code.

Have the submitter script invoke the Deadline GUI bundle submitter instead.

Less code and dependencies. Installation is far simpler - just copying a single script to Keyshot's scripts directory. No special environment variables required.

- Added unit tests.
- Submitted jobs from Keyshot
- Verified that settings from previous submissions (parameters and attachments) are sticky and loaded back for the next submission *as long as* the submission has the same name.

Yes.

Not quite. This new submitter can work in place of the old one with no additional steps. It looks slightly different since it uses the bundle GUI submitter instead of the the custom submitter GUI we previously had. Also if job settings are sticky only if the job name stays the same (whereas previously the settings were associated with the scene file, not the submission name).

----


src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,386 @@
+# AUTHOR AWS
+# VERSION 0.0.6
```
lucaseck: nit: We don't have a great way to ensure this gets updated right now, but can it be bumped up to 0.1.3 which is the current latest

src/deadline/keyshot_submitter/default_keyshot_job_template.yaml
lucaseck, line 1: Not as part of this PR, but in general I think this should be moved to https://github.com/aws-deadline/deadline-cloud-samples as an example

crowecawcaw: Makes sense. I'll leave it for the moment but can follow up.

README.md
```diff
@@ -27,15 +27,9 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 
 1. Install deadline and Pyside6
     - e.g. `pip install deadline[gui]`
-2. Copy or link the file `deadline-cloud-for-keyshot/keyshot_script/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder.
+2. Copy or link the file `deadline-cloud-for-keyshot/keyshot_submitter/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder.
     - e.g. On Windows `C:/Users/<USER>/Documents/KeyShot/Scripts`
```
lucaseck: nit: Can this be updated to reflect both local and system install locations like we are doing down below

install_builder/deadline-cloud-for-keyshot.xml
```diff
@@ -117,7 +95,6 @@
             <value>Deadline Cloud for KeyShot
 - Install the integrated KeyShot submitter files to the installation directory
```
lucaseck: `Install the integrated KeyShot submitter files to the installation directory` should be removed as well since the only file being "installed" is the `Submit to AWS Deadline Cloud` script mentioned on the line below.

pyproject.toml
lucaseck, line 1: "deadline == 0.48.*" will need to be bumped. Interesting case though where `0.48.5` is the minimum required, so we will likely need to temporarily lock to `"deadline >= 0.48.5,< 0.49"` or something similar

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,386 @@
...
+RENDER_SUBMITTER_SETTINGS_FILE_EXT = ".deadline_render_settings.json"
+
+
+@dataclass
+class Settings:
+    scene_file: str
+    frames: str
+    output_file_path: str
+    output_format: str
+    input_filenames: list[str]
```
lucaseck, line 24: Normally any of the JA lists are stored as sets, is there a reason I'm missing why these need to be lists?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,386 @@
...
+                "description": "The frames to render. E.g. 1-3,8,11-15",
+                "minLength": 1,
+            },
+            {
+                "name": "OutputFilePath",
+                "type": "PATH",
+                "objectType": "FILE",
+                "dataFlow": "OUT",
+                "userInterface": {
+                    "control": "CHOOSE_INPUT_FILE",
```
lucaseck: I believe `CHOOSE_OUTPUT_FILE` exists. Should that be used instead?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,386 @@
...
+                            "cancelation": {"mode": "NOTIFY_THEN_TERMINATE"},
+                        }
+                    },
+                },
+            }
+        ],
+    }
+
+
+def construct_asset_references(settings: Settings) -> dict:
```
lucaseck, line 250: nit: return value typing can be expanded, looks like there are a few other places as well that I won't leave a comment on each one.

install_builder/deadline-cloud-for-keyshot.xml
joel-wong-aws: Updated!

README.md
joel-wong-aws: Updated!

pyproject.toml
joel-wong-aws: Updated to `deadline >= 0.48.5`. I'm not sure why `<0.49` would be needed?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: These are saved as JSON, which doesn't support sets, so we use a list instead

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: Yup!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: I don't think expanding the return type adds much value here, but I've added docstrings indicating that the return values are json-convertible!

pyproject.toml
lucaseck: Will likely have to be bumped again from the pull of 0.48.5. We normally pin to 0.x.* so I believe just pinning to >= 0.48.5 would allow 0.49 etc automatically which right now we don't want to happen, each minor version increment we want to be in control of.

README.md
```diff
@@ -27,15 +27,10 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 
 1. Install deadline and Pyside6
```
lucaseck: Should we add a note here that deadline should be > 0.48.7 for sticky settings to work at least? The update to the pyproject.toml won't enforce anything for the submitter, only the adaptor, so it's up to the user to have a new enough version of deadline present for sticky settings to work.

joel-wong-aws: Added a note and updated the recommended `pip install` instruction!

README.md
```diff
@@ -25,17 +25,12 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
...
-2. Copy or link the file `deadline-cloud-for-keyshot/keyshot_script/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder.
-    - e.g. On Windows `C:/Users/<USER>/Documents/KeyShot/Scripts`
-3. Set the following environment variables
-    - Set the environment variable `DEADLINE_PYTHON` as the path to the Python installation where deadline-cloud and PySide2 were installed in step 1.
-      - e.g. On Windows if using Python 3.10 it might be `set DEADLINE_PYTHON=C:/Users/<USER>/AppData/Local/Programs/Python/Python310/python`
-    - Set the environment variable `DEADLINE_KEYSHOT` as the path to the `<PATH TO>/deadline-cloud-for-keyshot/src/deadline/keyshot_submitter` folder
-      - e.g. On Windows if the source was on the user's desktop it might be  `set DEADLINE_KEYSHOT=C:/Users/<USER>/Desktop/deadline-cloud-for-keyshot/src/deadline/keyshot_submitter`
-4. Launch KeyShot with the environment variables from step 3. set.
-5. The submitter can be launched within KeyShot from `Window > Scripting Console > Scripts > Submit to AWS Deadline Cloud > Run`
+1. Install `deadline[gui] >= 0.48.7,< 0.49` and `PySide6`
```
lucaseck: nit: For the submitter I don't think we need to recommend pinning to < 0.49. We will have to update that every time and it's not required IMO.

README.md
```diff
@@ -25,17 +25,12 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
...
+    - e.g. `pip install "deadline[gui] >= 0.48.7,< 0.49" PySide6`
```
epmog: why specify PySide6 in the command? `deadline[gui]` already brings the version of `Pyside6-Essentials` that it requires. Whereas just PySide6 will bring a lot of additional addons.

Should probably just specify installing the keyshot package to ensure we get a compatible deadline version without needing to update this all the time. Can still specify deadline[gui] to pull in that optional dep.

`pip install deadline-cloud-for-keyshot "deadline[gui]"`

should do the trick

README.md
joel-wong-aws: I'm confused. Didn't we want to pin this based on an [earlier discussion](https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/93#discussion_r1655374540)?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,417 @@
...
+                            "embeddedFiles": [
+                                {
+                                    "name": "initData",
+                                    "filename": "init-data.yaml",
+                                    "type": "TEXT",
+                                    "data": (
+                                        "scene_file: '{{Param.KeyShotFile}}'\n"
+                                        "output_file_path: '{{Param.OutputFilePath}}'\n"
+                                        "output_format: 'RENDER_OUTPUT_{{Param.OutputFormat}}'\n"
+                                        ""
```
epmog: super nit: this line can be deleted.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,417 @@
...
+                capture_output=True,
+                text=True,
+            )
+            output = result.stdout
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"AWS Deadline Cloud KeyShot submitter could not open: {e.stderr}")
+    try:
+        return json.loads(output)
+    except json.JSONDecodeError as e:
+        print(f"Error parsing settings file: {e}")
```
epmog: copy/paste? Should indicate it failed to read deadline-cli output as json

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -0,0 +1,417 @@
...
+        return None
+
+
+def main():
+    scene_file = lux.getSceneInfo()["file"]
```
epmog, line 361: Do we need to check if the scene has been saved first? This might not exist, or other settings may fail to be set?

README.md
joel-wong-aws: Removed PySide6 from the `pip install` command. Is `pip install deadline-cloud-for-keyshot` necessary? If someone is running these instructions, I understand that they would be doing some sort of local development and having an install of `deadline-cloud-for-keyshot` may cause more confusion around import/path priority? If someone doesn't want to do development of the KeyShot submitter, I understand that they would just install it from the Deadline Cloud binaries following our public docs. We could mention that here as well once we release the latest binaries.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: Removed!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: Updated!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
joel-wong-aws: Out of scope, the existing implementation also assumes that the file exists

README.md
epmog: Gotcha. In my mind this was related to the manual installation instructions, rather than solely the development instructions. Looking at it right now, there's nothing to indicate that it's development, manual installation, or a continuation from the submitter installer - the section is just "Using the submitter".

Future improvement we can punt - but we'll want to make this clear.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
epmog: 😬 fair enough - resolving.

epmog (dismissed):

README.md
epmog: > Is pip install deadline-cloud-for-keyshot necessary?

It depends on what the instructions are targeting. Doing this gets you the correct version of deadline which will then mean we don't have to keep updating these instructions on new releases.


1. If you're doing a local build, `pip install`ing the built wheel will grab you the correct version.
2. If you're doing a manual install, pip installing from pypi gets you the right dependency versions for the installed version
3. If you're continuing from a submitter installer build none of this makes sense.

README.md
lucaseck: Spoke offline, and this can be punted until after merging. No sense holding this PR up more.

lucaseck (dismissed):

lucaseck (approved):

epmog (approved):

PR #97, opened 2024-07-05
feat: support ksp bundling

lucaseck:
There was no pathmapping support with KeyShot. 
One way around this is to bundle a scene up as a KeyShot Package(ksp) which creates a zipped archive of the scene file and any external file references. The scene within is modified to have relative paths to all of the referenced files. 

This PR adds a submission mode selector for:
1. Just submit the scene .bip file and all files referenced within will have to be available through network storage or some other means
2. Automatically save the current scene as a ksp to a temp directory, unpack it and then submit the modified scene and assets from the unpacked ksp where all paths are relative
KeyShot can be used without having to deal with manually packing/unpacking ksps or setting up network storage
- Ran all unit tests including newly included ones
- Ran submissions using both submission modes and verified the outputs
- Tested canceling the dialog, and the sticky setting of which option was selected
Only in the README
No
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,24 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
         - You can navigate to the folder by going to Finder, clicking the menu for Go -> Go to Folder, and typing in the folder path.
 3. Launch KeyShot. The submitter can be launched within KeyShot from `Window > Scripting Console > Scripts > Submit to AWS Deadline Cloud > Run`
 
+#### Submission Modes
+
+There are two submission modes for the KeyShot submitter which a dialog will ask you to select from before opening the submitter UI.
+
+1. Attach `The scene BIP file and all external files references`
+    - The open scene file and all external files referenced within will be included
+    as job attachments. The submitter will export the open scene to a
+    KSP(KeyShot Package) which turns all file paths in the scene into relative
+    paths and creates a flattened directory with all of the external files directly
+    beside the scene file. The KSP is then unzipped, and the new scene file with
+    the relative paths and all external files will be submitted with the job.
+    The temporary directory used to save the KSP will be deleted after each
+    submission.
+1. Attach `Only the scene BIP file`
+    - Only the open scene file will be attached to the submission. The expectation is that any
+    external files referenced within the scene will be available to the workers
+    through network storage or some other method.
+
 ## Adaptor
 
 The KeyShot Adaptor implements the [OpenJD][openjd-adaptor-runtime] interface that allows render workloads to launch KeyShot and feed it commands. This gives the following benefits:
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -7,15 +7,20 @@ import json
 import os
 import platform
 import shlex
+import shutil
 import subprocess
 import tempfile
+import glob
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Optional, Tuple
 
 import lux
 
 RENDER_SUBMITTER_SETTINGS_FILE_EXT = ".deadline_render_settings.json"
+SUBMISSION_MODE_KEY = "submission_mode"
+# Unique ID required to allow KeyShot to save selections for a dialog
+DEADLINE_CLOUD_DIALOG_ID = "e309ce79-3ee8-446a-8308-10d16dfcbb42"
 
 
 @dataclass
@@ -26,7 +31,6 @@ class Settings:
     output_directories: list[str]
     referenced_paths: list[str]
     auto_detected_input_filenames: list[str]
-    auto_detected_output_directories: list[str]
 
     def output_sticky_settings(self):
         return {
@@ -88,11 +92,7 @@ class Settings:
         if asset_references.get("inputs", {}).get("directories"):
             self.input_directories = asset_references["inputs"]["directories"]
         if asset_references.get("outputs", {}).get("directories"):
-            # Persist output directories that were not autodetected (i.e. were manually added)
-            self.output_directories = list(
-                set(asset_references["outputs"]["directories"])
-                - set(self.auto_detected_output_directories)
-            )
+            self.output_directories = asset_references["outputs"]["directories"]
         if asset_references.get("referencedPaths"):
             self.referenced_paths = asset_references["referencedPaths"]
 
@@ -259,18 +259,7 @@ def construct_asset_references(settings: Settings) -> dict:
                     list(set([*settings.input_filenames, *settings.auto_detected_input_filenames]))
                 ),
             },
-            "outputs": {
-                "directories": sorted(
-                    list(
-                        set(
-                            [
-                                *settings.output_directories,
-                                *settings.auto_detected_output_directories,
-                            ]
-                        )
-                    )
-                ),
-            },
+            "outputs": {"directories": sorted(settings.output_directories)},
             "referencedPaths": sorted(settings.referenced_paths),
         }
     }
@@ -346,6 +335,7 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
                 check=True,
                 capture_output=True,
                 text=True,
+                creationflags=subprocess.CREATE_NO_WINDOW,  # type: ignore[attr-defined]
             )
             output = result.stdout
     except subprocess.CalledProcessError as e:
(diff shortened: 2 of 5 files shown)


src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -357,55 +346,137 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
...
+    input_filenames = [
+        os.path.join(unpack_dir, file)
+        for file in os.listdir(unpack_dir)
+        if not file.endswith(".bip")
+    ]
+
+    bip_files = glob.glob(os.path.join(unpack_dir, "*.bip"))
+
+    if not bip_files:
+        raise RuntimeError("No .bip files found in the KSP bundle.")
```
joel-wong-aws, line 416: Should we raise an error if there are multiple bip files?

test/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -242,3 +243,45 @@ def test_settings_apply_submitter_settings():
...
     assert sorted(settings.output_directories) == ["test_directory_2", "test_directory_4"]
     assert sorted(settings.referenced_paths) == ["test_ref_path_2"]
+
+
+def test_save_ksp_bundle():
+    dir = os.path.normpath("/testdir/test")
+    bundle_name = "test_bundle.ksp"
+    expected_bundle_path = os.path.normpath(f"{dir}/{bundle_name}")
+
+    output = submitter._save_ksp_bundle(dir, bundle_name)
```
joel-wong-aws: nit: if we're testing this function, should it be a public function (without leading underscore)?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -357,55 +346,137 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
...
+    Returns a dictionary of the selected option values in the format:
+        {'SUBMISSION_MODE_KEY': [1, 'only the scene BIP file']}
+    """
+    dialog_items = [
+        (
+            SUBMISSION_MODE_KEY,
+            lux.DIALOG_ITEM,
+            "Do you want to submit",
+            0,
+            ["the scene BIP file and all external files references", "only the scene BIP file"],
```
crowecawcaw: I think the wording could be tweaked a bit so an artist would know how to choose one setting versus another. Maybe include a description like:
> Choose "the scene BIP file and all external file references" if you're rendering on service-managed fleets or to upload your scene to a customer-managed fleet. Choose "only the scene BIP file" if you're rendering on customer-managed fleets that already have access to the file references, for example on a shared network drive. 

Nit: Also instead of "Do you want to submit", maybe rephrase as "What files do you want to attach with your job?" That way the user selection answers the question instead of form part of the question itself.

test/keyshot_submitter/test_keyshot_submitter.py
lucaseck: Leftover from before my initial rebase. Removed the leading `_` from all new functions to follow the pattern we've already established in the file.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
lucaseck: I'm not sure that there is ever a case where there can be multiple BIP files, but I've added the error just in case.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
lucaseck: There's nothing stopping someone from setting up network storage for service managed fleets as far as I know. I don't think we can make the suggestion of which to use based on if SMF or CMF are being used.

lucaseck: Updated the wording of the prompt itself

crowecawcaw (dismissed):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -370,54 +436,62 @@ def main(lux):
...
-    job_template = construct_job_template(filename)
-    asset_references = construct_asset_references(settings)
-    parameter_values = construct_parameter_values(settings)
+    with tempfile.TemporaryDirectory() as bundle_temp_dir:
+        # {'submission_mode': [0, 'the scene BIP file and all external files references']}
+        if not dialog_selections[SUBMISSION_MODE_KEY][0]:
+            temp_scene_file, input_filenames = get_ksp_bundle_files(bundle_temp_dir)
+            settings.auto_detected_input_filenames = input_filenames
+            settings.parameter_values.append({"name": "KeyShotFile", "value": temp_scene_file})
+        else:
```
crowecawcaw, line 483: Do we need to attach the scene file in this branch? Also do we still want to attach external files if KSP bundling isn't selected? (I think not, but just double checking.)

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -370,54 +436,62 @@ def main(lux):
         else:
             lux.saveFile()
 
-    scene_file = lux.getSceneInfo()["file"]
-    external_files = lux.getExternalFiles()
+    dialog_selections = options_dialog()
+
+    if not dialog_selections:
+        print("Job submission canceled.")
+        return
```
crowecawcaw, line 443: We should eventually raise an exception here so the job status doesn't show Suceess afterwards. Related PR: https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/102 Whichever PR get's merged first, we should fix the other.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
lucaseck: The scene file parameter is a PATH parameter, just by having it specified it will include it as an input file, no need to actually add it to the asset references directly

lucaseck: No, if just the .bip file is selected we don't want to attach any of the external files since they should be made available through network storage or some other means.

crowecawcaw (approved):

README.md
```diff
@@ -34,6 +34,24 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
...
+There are two submission modes for the KeyShot submitter which a dialog will ask you to select from before opening the submitter UI.
+
+1. Attach `The scene BIP file and all external files references`
+    - The open scene file and all external files referenced within will be included
+    as job attachments. The submitter will export the open scene to a
+    KSP(KeyShot Package) which turns all file paths in the scene into relative
+    paths and creates a flattened directory with all of the external files directly
+    beside the scene file. The KSP is then unzipped, and the new scene file with
+    the relative paths and all external files will be submitted with the job.
+    The temporary directory used to save the KSP will be deleted after each
```
ryyakobe, line 48: Have we considered not deleting this temp KSP file but saving it with submission timestamp, for example? This way we would create a "snapshot" of the scene for each submission, so it would be easy to jump back to the required revision, if needed. Not sure if there is a real use-case behind it.

Also, if current scene is actually a part of existing KSP file, it seems like we would still create a temp SKP file, update all the paths, create temp directory and delete it. Do we need to have all these steps for existing KSP files?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -357,7 +347,83 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
...
+                  KSP bundle before submission.
+    Returns a dictionary of the selected option values in the format:
+        {'SUBMISSION_MODE_KEY': [1, 'only the scene BIP file']}
+    """
+    dialog_items = [
+        (
+            SUBMISSION_MODE_KEY,
+            lux.DIALOG_ITEM,
+            "What files would you like to attach to the job?",
+            0,
```
ryyakobe, line 366: What does `0` mean here? Is it the default selected "index" in `["The scene BIP file and all external files references", "Only the scene BIP file"]`?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -357,7 +347,83 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
...
+
+def save_ksp_bundle(directory: str, bundle_name: str) -> str:
+    """
+    Saves out the current scene and any file references to a ksp bundle in a
+    directory.
+    Returns the file path where the bundle was saved to.
+    """
+    full_ksp_path = os.path.join(directory, bundle_name)
+
+    if not lux.savePackage(path=full_ksp_path):
```
ryyakobe, line 387: What happens if the file with this name already exists? Would we override it or would we get an error?

What do you think about checking if the file already exists before trying to save it? Does it make sense to add some randomness, like timestamp, to the `temp_deadline_cloud.ksp` filename to avoid name collisions, even if the temp file has not been cleaned up properly? Similar for `unpack` directory.

ryyakobe (approved):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
lucaseck: 0 is just the index of the options, default to the first option which is the scene file and all external references

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
lucaseck: It's being created in a temp directory that is deleted at the end of every submission, there should never be any name collisions.

README.md
lucaseck: I considered it, but couldn't think of a good enough reason right now to add the extra complexity of managing that. No issue with it being a future improvement if that's a workflow that's requested.

As far as I know if a user tries to open an existing ksp up it has to be extracted by KeyShot somewhere(either the Resources folder or another location) before KeyShot can be used, so everything should be as a bip file during submission

PR #101, opened 2024-07-10
feat: submitter prompts user to save changes before submitting

crowecawcaw:
If the user makes updates to their scene but fails to save them before submitting, their local changes won't be part of the submission.

Detect when there are unsaved changes and prompt the user to save their file first.

Screenshot 2024-07-10 at 2 29 05 PM

Local changes are always saved before submitting.

Local testing and a unit test.

No, not needed.

No
----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -359,11 +359,16 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
 
 def main(lux):
     if lux.isSceneChanged():
-        lux.getMessageBox(
-            title="Unsaved changes", msg="You have unsaved changes. Save your scene and try again."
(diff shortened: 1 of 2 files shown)


src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -357,7 +357,14 @@ def gui_submit(bundle_directory: str) -> Optional[dict[str, Any]]:
         return None
 
 
-def main():
+def main(lux):
+    if lux.isSceneChanged():
+        lux.getMessageBox(
+            title="Unsaved changes", msg="You have unsaved changes. Save your scene and try again."
```
lucaseck: Can we present this more as a "Unsaved changes detected, would you like to save now?" option that calls lux.SaveFile() if the user chooses to save first and then continues submission?

lucaseck (approved):

epmog (approved):

PR #102, opened 2024-07-10
fix: do not show Success status when submission is cancelled

crowecawcaw:
When the submitter window is cancelled, Keyshot pops up a result window saying the script was "Successful" which is confusing/.

Make the status show as "Failure" instead. "Cancelled" would be a better status, but we can't customize it and it's less confusing than "Success".

Screenshot 2024-07-10 at 4 05 08 PM

I cancelled a job.

No, not needed

No
----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -439,8 +439,8 @@ def main(lux):
     dialog_selections = options_dialog()
 
     if not dialog_selections:
-        # Dialog was cancelled. Raise an exception so Keyshot does not show the script's result status as "Success"
-        raise Exception("Job submission canceled.")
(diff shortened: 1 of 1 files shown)


lucaseck (dismissed):

diane-hj (dismissed):

karthikbekalp (dismissed):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -494,6 +494,10 @@ def main(lux):
         output = gui_submit(bundle_temp_dir)
 
     if output:
+        if output.get("status") == "CANCELED":
+            # Raise an exception so Keyshot does not show the script's result status as "Success"
+            raise Exception("Submission was cancelled.")
```
diane-hj: Nit. Looks like the spellings of `canceled` and `cancelled` should match :)

diane-hj (approved):

lucaseck (approved):

Issue #107, opened 2024-08-22
Bug: Missing python libs in KeyShot 13


hypnoboutique:
### Expected Behaviour

Works

### Current Behaviour

Does not work

```
Traceback (most recent call last):
  File "<string>", line 9, in <module>
ModuleNotFoundError: No module named 'shlex'
```

![Screenshot 2024-08-22 at 09 58 55](https://github.com/user-attachments/assets/b448e636-cef0-405e-be73-a16e9256ae0d)

### Reproduction Steps

- Install Keyshot 13 on mac OS 14.5
- Follow (incorrect) install steps [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot) (the script is not located at `deadline-cloud-for-keyshot/keyshot_submitter/Submit to AWS Deadline Cloud.py`, it is in `deadline-cloud-for-keyshot/src/deadline/keyshot_submitter/`)
- Attempt to submit from scripts panel in Keyshot

A number of python modules are missing in the Keyshot bundled version of python. Initially, as seen in my screenshot, this was shlex. I copied shlex, pathlib, zipfile from a local dist of python and was able to make something happen but still hit an error eventually because zlib wasn't available.

### Code Snippet

n/a

hypnoboutique:
I finally got this script to complete by patching the missing libs and modules, modified my local version and shell profile (temporarily, I can't leave them like this). Now, after creating a job, every task fails with:

```
2024/08/22 14:28:30+01:00  
2024/08/22 14:28:30+01:00 ==============================================
2024/08/22 14:28:30+01:00 --------- Entering Environment: KeyShot
2024/08/22 14:28:30+01:00 ==============================================
2024/08/22 14:28:30+01:00 ----------------------------------------------
2024/08/22 14:28:30+01:00 Phase: Setup
2024/08/22 14:28:30+01:00 ----------------------------------------------
2024/08/22 14:28:30+01:00 Writing embedded files for Environment to disk.
2024/08/22 14:28:30+01:00 Mapping: Env.File.initData -> /sessions/session-7f2614ae86654595964452509e2687c5bx5_o9qe/embedded_filessp_jmspe/init-data.yaml
2024/08/22 14:28:30+01:00 Wrote: initData -> /sessions/session-7f2614ae86654595964452509e2687c5bx5_o9qe/embedded_filessp_jmspe/init-data.yaml
2024/08/22 14:28:30+01:00 ----------------------------------------------
[...]

PR #108, opened 2024-08-26
chore: update README with latest KeyShot submitter location

joel-wong-aws:
The KeyShot submitter location was incorrect in the README.

Addresses the README issue in https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/107

Updated the README. Also updated the Windows scripts location described in the README to be more open/version compatible.

Correct installation instructions.

N/A

N/A - this is documentation.

No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,10 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 ### Installing and using the KeyShot Submitter
 
 1. Run `pip install deadline[gui]`
-2. Copy the file `deadline-cloud-for-keyshot/keyshot_submitter/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder for your OS:
+2. Copy the file `deadline-cloud-for-keyshot/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder for your OS:
     - Windows (choose one): 
-        - User scripts folder: `%USERPROFILE%/Documents/KeyShot/Scripts`
-        - System-wide scripts folder: `%PROGRAMFILES%/KeyShot/Scripts`
+        - User scripts folder e.g. `%USERPROFILE%/Documents/KeyShot/Scripts`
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):

README.md
```diff
@@ -26,10 +26,10 @@ This package provides a KeyShot plugin script that creates jobs for AWS Deadline
 ### Installing and using the KeyShot Submitter
 
 1. Run `pip install deadline[gui]`
-2. Copy the file `deadline-cloud-for-keyshot/keyshot_submitter/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder for your OS:
+2. Copy the file `deadline-cloud-for-keyshot/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py` to the KeyShot scripts folder for your OS:
```
epmog, line 29: Are we sure that the instructions are actually wrong? 

~~This is now assuming you're git cloning, but the instructions right above say `pip install deadline[gui]`. So I'm thinking that there's two potential workflows getting criss-crossed. There's a developer workflow with a local checkout of keyshot, and then there's a user flow that installing releases.~~

Edit: Slightly inaccurate in my above comment, I'll correct it. There should be two workflows, 1 aimed at users and 1 aimed at developers. With the updated instructions, this now requires a git clone of the repo which doesn't feel right for a user.

So my suggestion is: README.md should probably have user instructions that leverage `pip install deadline-cloud-for-keyshot "deadline[gui]"`, and the DEVELOPER.md should probably have developer instructions with a local checkout of keyshot. Thoughts?

joel-wong-aws: Because `deadline-cloud-for-keyshot` has not been released since June 19th, `pip install deadline-cloud-for-keyshot "deadline[gui]"` does not yet contain the latest version of the submitter and the instructions would be incorrect. That's why we still have these additional instructions for both users and developers. When the next release of `deadline-cloud-for-keyshot` occurs, I think we can update these instructions.

epmog: right, sounds like we need to release soon ;)

epmog (approved):

PR #105, opened 2024-08-07
chore: add script to create adaptor packaging artifact

joel-wong-aws:
### What was the problem/requirement? (What/Why)

Create a script to produce the adaptor packaging artifact for KeyShot, similar to those for [Blender](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/scripts/create_adaptor_packaging_artifact.sh), [Houdini](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/mainline/scripts/create_adaptor_packaging_artifact.sh), and [Nuke](https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/scripts/create_adaptor_packaging_artifact.sh)

### Was this change documented?

No, N/A

### Is this a breaking change?

No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #109, opened 2024-08-27
chore: create adaptor packaging artifact script

crowecawcaw:
We need a script to package up this adapter as part of the conda build process.

Make a packaging script. Most code was pulled from [this PR](https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/105) but with some tweaks to clean up some bits that did not end up being used and to make pywin32 portable.

We can build a conda package for the Keyshot adapter.

I built the artifact, built a conda package from it, and then ran a render using the adapter.

Not needed.

No
----

### DIFF
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py
@@ -21,6 +21,8 @@ if "openjd" not in sys.modules.keys():
                 print("add_dll_directory failed: %s" % p)
         sys.path.append(p)
 
+# Required for making pywin32 portable
+import pywin32_bootstrap  # type: ignore # noqa: F401 E402
 
 from types import FrameType  # noqa: E402
 from typing import Optional  # noqa: E402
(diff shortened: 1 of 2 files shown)


src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py
```diff
@@ -21,6 +21,8 @@
                 print("add_dll_directory failed: %s" % p)
         sys.path.append(p)
 
+# Required for making pywin32 portable
+import pywin32_bootstrap  # type: ignore # noqa: F401 E402
```
lucaseck, line 25: Any link or paper trail for why this is necessary?

crowecawcaw: Good call. There's some info here: https://github.com/mhammond/pywin32/blob/main/win32/Lib/pywin32_bootstrap.py#L1-L8

Usually it gets run automatically as the module is loaded by a `.pth`, but `.pth` files are only run when modules are in the site-packages folder from what I've read.

lucaseck (dismissed):

epmog (approved):

lucaseck (approved):

PR #110, opened 2024-08-28
fix: add default values for conda packages and channels

crowecawcaw:
The CLI-based submitter should provide defaults for conda if the queue has a conda queue environment,

Add defaults.

The submitter works easily with conda without fiddling with conda settings.

I submitted a job that used the defaults to a queue that had the default queue environment and it worked. I also tested it with a queue that did not have a queue environment.

Not needed.

No, it adds back a feature that's been missing since the submitter refactor. Note: it defaults to a different Keyshot version scheme that the old defaults.
----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -448,7 +448,6 @@ def get_ksp_bundle_files(directory: str) -> Tuple[str, list[str]]:
 
 
 def main(lux):
-
     if lux.isSceneChanged():
         result = lux.getInputDialog(
             title="Unsaved changes",
@@ -509,8 +508,9 @@ def main(lux):
             settings.parameter_values.append({"name": "KeyShotFile", "value": scene_file})
 
         # Add default values for Conda
+        major_version, minor_version = lux.getKeyShotDisplayVersion()
         settings.parameter_values.append(
-            {"name": "CondaPackages", "value": "keyshot=2024.* keyshot-openjd=0.1.*"}
+            {"name": "CondaPackages", "value": f"keyshot={major_version}.* keyshot-openjd=0.1.*"}
         )
         settings.parameter_values.append({"name": "CondaChannels", "value": "deadline-cloud"})
 
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -55,9 +55,11 @@ class Settings:
 
         for param in input_parameter_values:
             if param.get("name") and param.get("value"):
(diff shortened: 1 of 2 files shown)


src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -483,6 +485,12 @@ def main(lux):
         else:
             settings.parameter_values.append({"name": "KeyShotFile", "value": scene_file})
 
+        # Add default values for Conda
+        settings.parameter_values.append(
+            {"name": "CondaPackages", "value": "keyshot=2024.* keyshot-openjd=0.1.*"}
```
lucaseck: Can we just `lux.getKeyShotDisplayVersion()` to autopopulate?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -55,9 +55,11 @@ def apply_sticky_settings(self, sticky_settings: dict):
 
         for param in input_parameter_values:
             if param.get("name") and param.get("value"):
-                # don't re-use KeyShotFile param if the render settings are copied from one file to another
-                if param["name"] != "KeyShotFile":
-                    updated_parameter_values[param["name"]] = param["value"]
+                # Don't re-use KeyShotFile param if the render settings are copied from one file to another
+                # Don't preserve conda settings so the Keyshot version can be updated by updating the submitter
+                if param["name"] in ["KeyShotFile", "CondaPackages", "CondaChannels"]:
```
joel-wong-aws, line 57: Why aren't we using sticky settings for `CondaChannels`? Customers may want to add another Conda channel (e.g. `conda-forge`) in addition to the `deadline-cloud` channel, and the `deadline-cloud` channel name won't change, right?

I also think we could add some extra logic around `CondaPackages` to ensure we have the correct `keyshot` and `keyshot-openjd` versions but retain any additional packages. But that's out of scope for this PR, customers also always have the option to customize the submitter 🙂

crowecawcaw: The use case I had in mind was if someone had a scene in Keyshot 2024 and then updated to Keyshot 2025, their conda packages parameter should adjust to match instead of pulling from the last submission.

If we detected if someone had edited the channels, that would cover both cases but I don't think that's worth the complexity, especially if someone can edit the plugin themselves like you pointed out.

Those are my thoughts but I'm open to changing it.

joel-wong-aws: I agree that because customers can always update this themselves, it's not a blocker for this PR and this is a reasonable default.

joel-wong-aws (approved):

lucaseck (approved):

PR #111, opened 2024-08-28
fix: remove submitter dependencies that aren't available in KeyShot Python

joel-wong-aws:
Addresses https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/107

Replaced each of the unavailable libraries with alternate approaches.

Note that any OS which can run KeyShot 2023 or 2024 can run `unzip` (Mac) or `Expand-Archive` (Windows).

Customers are able to submit in KeyShot Python.

Uninstalled system Python 3.11 from my (Windows) machine. Submitted jobs with and without the external file options.
Also tested from a Mac machine.

No, not required.

No

----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -6,13 +6,10 @@
 import json
 import os
 import platform
-import shlex
-import shutil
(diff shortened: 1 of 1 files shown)


crowecawcaw (approved):
Worked on a Mac with and without KSP bundling

epmog (approved):

PR #123, opened 2024-10-09
fix: update installer to locate KeyShot scripts directory in more cases

crowecawcaw:
The KeyShot installer didn't install the submitter script in some circumstances. Specifically, I had a KeyShot install where the KeyShot resources folder pointed to a directory that did not yet exist.

- Create the scripts directory if it doesn't exist already
- Use a fallback value for the scripts directory if one cannot be found from environment variables (e.g. if someone runs the Deadline installer before running the KeyShot installer)

KeyShot installer works in more cases.

I ran the installer with KeyShot was installed and without, and in both cases it installed the plugin correctly.

Not needed

No
----

### DIFF
--- a/install_builder/deadline-cloud-for-keyshot.xml
+++ b/install_builder/deadline-cloud-for-keyshot.xml
@@ -53,7 +53,7 @@
             </conditionRuleList>
             <actionList>
                 <!-- KeyShot 2024 (and possibly beyond) env var -->
-                <setInstallerVariable name="keyshot_resources_directory" value="${env(KEYSHOT)}"/>
+                <setInstallerVariable name="keyshot_scripts_directory" value="${env(KEYSHOT)}\Scripts"/>
             </actionList>
             <elseActionList>
                 <if>
@@ -66,29 +66,18 @@
                     </conditionRuleList>
                     <actionList>
                         <!-- KeyShot 2023 env var -->
-                        <setInstallerVariable name="keyshot_resources_directory" value="${env(KEYSHOT12)}"/>
+                        <setInstallerVariable name="keyshot_scripts_directory" value="${env(KEYSHOT12)}\Scripts"/>
                     </actionList>
                     <elseActionList>
-                        <setInstallerVariable name="keyshot_resources_directory" value=""/>
+                        <!-- KeyShot 2024 default location as a fallback -->
+                        <setInstallerVariable name="keyshot_scripts_directory" value="${user_home_directory}\Documents\KeyShot\Scripts"/>
                     </elseActionList>
                 </if>
             </elseActionList>
         </if>
-        <if>
-            <conditionRuleList>
-                <compareText>
-                    <text>${keyshot_resources_directory}</text>
-                    <logic>does_not_equal</logic>
-                    <value></value>
-                </compareText>
-            </conditionRuleList>
-            <actionList>
-                <setInstallerVariable name="keyshot_scripts_directory" value="${keyshot_resources_directory}\Scripts"/>
-            </actionList>
-            <elseActionList>
-                <setInstallerVariable name="keyshot_scripts_directory" value=""/>
-            </elseActionList>
-        </if>
+        <createDirectory>
+            <path>${keyshot_scripts_directory}</path>
+        </createDirectory>
     </initializationActionList>
     <parameterList>
         <stringParameter name="deadline_cloud_for_keyshot_summary" ask="0" cliOptionShow="0">
--- a/install_builder/deadline-cloud-for-keyshot.xml
+++ b/install_builder/deadline-cloud-for-keyshot.xml
@@ -96,7 +96,7 @@
             <allowEmptyValue>0</allowEmptyValue>
             <ask>yes</ask>
             <cliOptionName>keyshot-scripts-directory</cliOptionName>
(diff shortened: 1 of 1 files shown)


install_builder/deadline-cloud-for-keyshot.xml
```diff
@@ -66,29 +66,18 @@
                     </conditionRuleList>
                     <actionList>
                         <!-- KeyShot 2023 env var -->
-                        <setInstallerVariable name="keyshot_resources_directory" value="${env(KEYSHOT12)}"/>
+                        <setInstallerVariable name="keyshot_scripts_directory" value="${env(KEYSHOT12)}\Scripts"/>
```
joel-wong-aws, line 69: I think I've reproduced the installer issue you've described. 
1. Download KeyShot Studio 2024.2 (also reproduced with KeyShot 2024.1 and KeyShot 2023.3)
2. Set the KeyShot resources directory to something different than the default `<user home>\Documents\KeyShot` (`<user home>\Documents\KeyShot12` for KeyShot 2023)

When this is done, the installer still sets the `KEYSHOT` env var to `<user home>\Documents\KeyShot` (for KeyShot 2023, the `KEYSHOT12` env var is set to `<user home>\Documents\KeyShot12`) even when the resources directory is elsewhere.

I agree that this behviour is confusing and we should work around it. However, I prefer the following approach:
1. Remove the `keyshot_resources_directory` variable as proposed in your PR and simplify it to just setting `keyshot_scripts_directory` directly (this is mostly a refactor)
2. Leave the scripts installer directory empty if neither env var is found (`<setInstallerVariable name="keyshot_scripts_directory" value="">` on line 73)
3. Instead of creating a folder at the default location during installer initialization if it doesn't exist (which could still potentially still be in the wrong spot), we should validate that the folder exists, and if it doesn't exist, force the user to input it (i.e. remove the `createDirectory` command proposed in this PR and set the `keyshot_scripts_directory` to empty if it doesn't exist)

lucaseck (commented):
I think we could simplify this for ourselves and keep the current default of using KEYSHOT/KEYSHOT12 to point to the scripts folder, but add a prompt in the installer to confirm the installation directory and let users pick their own if the default isn't correct? Any opinions on that path? It's called out in our [docs](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/submitter.html) that the DCCs should be installed before running the submitter installer so I think we can make the assumption that KeyShot should exist already and not have to handle that case further than maybe printing a warning if we can't find the directory set or specified.

install_builder/deadline-cloud-for-keyshot.xml
```diff
@@ -66,28 +66,21 @@
                     </conditionRuleList>
                     <actionList>
                         <!-- KeyShot 2023 env var -->
-                        <setInstallerVariable name="keyshot_resources_directory" value="${env(KEYSHOT12)}"/>
+                        <setInstallerVariable name="keyshot_scripts_directory_default" value="${env(KEYSHOT12)}\Scripts"/>
```
joel-wong-aws: Should this be `keyshot_scripts_directory` instead of `keyshot_scripts_directory_default`?

install_builder/deadline-cloud-for-keyshot.xml
```diff
@@ -101,7 +94,6 @@
             <description>KeyShot Scripts Directory</description>
             <explanation>Path to scripts directory in the KeyShot resources directory. For easiest installation, KeyShot should be installed before installing Deadline Cloud for KeyShot.</explanation>
             <allowEmptyValue>0</allowEmptyValue>
-            <default>Will be detected from the KEYSHOT or KEYSHOT12 environment variables during installation. If neither exists and the KeyShot component is enabled, this must be input.</default>
```
joel-wong-aws, line 104: Why are we removing this? It is used in CLI help text. If we're worried about it no longer matching the existing functionality, can we update to describe that the env var is only used if the directory exists?

install_builder/deadline-cloud-for-keyshot.xml
crowecawcaw: Eek, yeah. Rebase issue.

install_builder/deadline-cloud-for-keyshot.xml
crowecawcaw: The default property is for the default variable value, not really help text though it might have appeared in CLI help in a convenient location. In some cases, this text was being stored as the value for the scripts directory and appeared inside the directory path text box which looked wrong. I'll move the info about the env variables to the CLI help text.

crowecawcaw: With the latest change, the CLI help text looks like this:
Screenshot 2024-10-10 at 10 36 12 AM

joel-wong-aws: While this is technically intended for default values, it's effectively only used for help text based on the installer definition.

I actually think customers will find it easier to understand with the former version. I find this new version to be confusing because it says that "The default value is detected from the KEYSHOT or KEYSHOT12 environment variables if present." but the default property is missing despite the env vars being present on the machine.

At a minimum, I think we need to clarify that the default value is determined during installation and won't be empty. That's something that isn't clear with an empty default value.

crowecawcaw: This default test now appears in the text box when a scripts directory isn't detected, so we can't reuse it in the same way we were before. I'll clarify the help text about the default being determined at runtime.

joel-wong-aws (approved):

lucaseck (approved):

PR #126, opened 2024-10-16
fix: downgrade python-semantic-release to fix changelog

crowecawcaw:
python-semantic-release 9.11 seems to break the changelog generation.

Downgrade to 9.9 which was working.

Fixes changelog generation.

9.9 previously worked.

Not needed

No
----


karthikbekalp (approved):
Approved! But does this mean we need to downgrade this across all our packages?

joel-wong-aws (approved):

PR #150, opened 2025-01-17
fix: use utf-8 encoding for file reads and writes to fix utf-8 encodi…

crowecawcaw:
Windows submissions fail when characters like `é` are in a file path.

Enforce utf-8 encoding.

Fixes submissions errors on Windows.

Not tested! Works on my Mac, but the original issue only appears on Windows. Looking for a teammate using Windows to verify the fix.

n/a

No
----


joel-wong-aws:
I have tested this on Windows! The updated code is able to submit correctly. Note that the renders still fail on Deadline Cloud service managed fleets as the adaptor code in the latest published Conda package does not yet handle special characters correctly.

joel-wong-aws (approved):

karthikbekalp (approved):
Thanks for adding the code! Hopefully, the adaptors work with the updated open-jd fix.

PR #157, opened 2025-01-24
RFC: docs: add conda recipe and GitHub action for adapter conda builds

crowecawcaw:
A recommended way for running the adapter is as a Conda package. The adapter conda package is available in SMF but it's difficult to build. 

- Add a conda recipe for building the adapter conda package. 
- Also add a GitHub workflow which builds it and allows it to be downloaded 

Easier to build an adapter conda package. Also it's a step towards automated tests and releases.

I ran the workflow, downloaded the package, uploaded it to a conda channel, and successfully ran a render with it.

Yes

No
----


karthikbekalp (approved):
Discussed some details offline, but the PR/RFC overall looks good to me.

crowecawcaw:
POC is interesting but we'll probably do something else for conda package builds.

PR #159, opened 2025-01-24
fix: pause rendering while submitting

### LINKED ISSUE #158: Bug: Pause Local Rendering when bringing up the submission window
### Describe the problem

Sometimes when working in resource intensive scenes, bringing up the submitter window freezes the entire application and I need to force quit. The root cause is the high CPU load of the scene + the CPU load of the script seems to make Keyshot stall indefinitely.

### Proposed Solution

Before bringing up the submitter window, the script pauses local rendering.

Optionally we can have it resume rendering when the window is closed.

### Example Use Cases

Any time you are working in a scene that takes up close to 100% of your CPU you would want this feature so you don't accidentally freeze the app.

crowecawcaw:
If KeyShot is rendering while the submitter opens, the workstation can slow down or crash.

Fixes: https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/158
Pause rendering while submitting.

Avoids slowdowns and crashes.

Manually opening the submitter and watching it pause and resume rendering. Also unit tests.

Not needed.

No

----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -457,103 +457,104 @@ def get_ksp_bundle_files(directory: str) -> Tuple[str, list[str]]:
 
 def main(lux):
     if lux.isPaused():
-        should_unpause_on_finish = False
+        # If rendering is already paused, run the function normally
+        main_inner(lux)
     else:
-        should_unpause_on_finish = True
-        lux.pause()
+        # If render is not already paused, pause while running then resume
+        try:
+            lux.pause()
+            main_inner(lux)
+        finally:
+            lux.unpause()
 
-    try:
-        if lux.isSceneChanged():
-            result = lux.getInputDialog(
-                title="Unsaved changes",
-                values=[
-                    (lux.DIALOG_LABEL, "You have unsaved changes. Do you want to save your file?")
-                ],
-            )
-            # result is {} if the user clicks Ok and None if the user clicks cancel
-            if result is None:
-                # Raise an exception so Keyshot shows the script's result status as "Failure" instead of "Success"
-                raise Exception("Changes must be saved before submitting.")
-            else:
-                lux.saveFile()
 
-        dialog_selections = options_dialog()
+def main_inner(lux):
+    if lux.isSceneChanged():
+        result = lux.getInputDialog(
+            title="Unsaved changes",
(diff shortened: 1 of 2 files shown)


src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -456,89 +456,104 @@ def get_ksp_bundle_files(directory: str) -> Tuple[str, list[str]]:
...
-                "value": "PNG",
-            },
-        ],
-        input_filenames=[],
-        auto_detected_input_filenames=[],
-        input_directories=[],
-        output_directories=[],
-        referenced_paths=[],
-    )
+    try:
```
crowecawcaw: The diff below here is just from whitespace changes. The handler is wrapped in a try-finally block. The finally block is the only new part.

joel-wong-aws: Instead of indenting this entire block in a `try-catch`, can we move it into a function? Or make the lux pausing/unpausing into a decorator? This has the benefit of improving code readability (fewer indents/conditions to keep in mind) and also reduces the diff (lower chance of merge conflicts)

joel-wong-aws (changes requested):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
crowecawcaw: Nice, went with the two functions. I tried playing with the decorator idea at first but it was looking more complex.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -456,6 +456,19 @@ def get_ksp_bundle_files(directory: str) -> Tuple[str, list[str]]:
 
 
 def main(lux):
+    if lux.isPaused():
+        # If rendering is already paused, run the function normally
+        main_inner(lux)
+    else:
```
joel-wong-aws, line 462: nit: to reduce indentation, we could use a `return` statement at the the above `if` statement
```
if lux.isPaused():
    # If rendering is already paused, run the function normally
    main_inner(lux)
    return

# If render is not already paused, pause while running then resume
try:
    lux.pause()
    main_inner(lux)
finally:
    lux.unpause()
```

joel-wong-aws (approved):

karthikbekalp (approved):

PR #178, opened 2025-02-10
chore: update python-semantic-release verion requirement

crowecawcaw:
python-semantic-release has frequent minor updates that trigger noisy dependabot PRs.

Loosen the version requirement.

Fewer dependabot PRs while still using the latest version of python-semantic-release.

Builds successfully

n/a

No

----

### DIFF
--- a/requirements-release.txt
+++ b/requirements-release.txt
@@ -1,1 +1,1 @@
-python-semantic-release == 9.18.*
+python-semantic-release == 9.*


karthikbekalp (approved):
Great thinking. I'll do the same for Cinema 4D as well.

joel-wong-aws (approved):

PR #184, opened 2025-02-27
test: add integration tests

joel-wong-aws:
To help avoid regressions when making changes to the KeyShot integration, we are adding integration tests.

`hatch run integ:test`. Note that there is one outstanding item around render precision/sampling. I will investigate this and updated in a separate PR.

Yes, I added more information to `DEVELOPMENT.md` about how to run these.

No

----


test/integ/helpers/test_runners.py
```diff
@@ -0,0 +1,194 @@
...
+
+def is_valid_template(template_location: Path) -> bool:
+    output = run_command(["openjd", "check", str(template_location), "--output", "json"])
+    output_json = json.loads(output.stdout)
+    return output_json["status"] == "success"
+
+
+def assert_expected_job_bundle_and_generated_job_bundle_are_equal(
+    expected_job_bundle_dir_path: Path, generated_job_bundle_dir_path: Path
+) -> None:
```
crowecawcaw, line 122: This would be great a common OJD or Deadline library function.

test/integ/helpers/submitter_test.py
```diff
@@ -0,0 +1,47 @@
...
+    # setup deadline imports
+    add_deadline_to_path()
+    # We can't use traditional imports due to spaces in the file name.
+    # The spaces are necessary because KeyShot uses the file name as the script name.
+    deadline = __import__("deadline.keyshot_submitter.Submit to AWS Deadline Cloud")
+    submitter = getattr(deadline.keyshot_submitter, "Submit to AWS Deadline Cloud")
+
+    # create the scene
+    lux.newScene()
+    scene_usda = os.path.join(scene_location, "scene.usda")
```
crowecawcaw, line 34: Why USD files instead of a bip?

test/integ/helpers/submitter_test.py
```diff
@@ -0,0 +1,47 @@
...
+    lux.importFile(path=scene_usda)
+    lux.saveFile(path=scene_usda.replace(".usda", ".bip"))  # scene.bip
+
+    # run the submitter
+    submitter.main(show_gui=False, export_dir=output_location)
+
```
karthikbekalp, line 40: Hmm, do we cleanup the test files created later on? Once the tests are complete?

hatch.toml
```diff
@@ -32,6 +32,15 @@ build = "hatch build"
...
 
+[envs.integ]
+pre-install-commands = [
+  "pip install -r requirements-integ-testing.txt",
+  "pip install -r requirements-testing.txt"
+]
+
+[envs.integ.scripts]
+test = "pytest --no-cov {args:test/integ} -vvv --numprocesses=1"
+
```
karthikbekalp, line 43: Hmm, how long do the submitter tests take vs the adaptor? If the time taken to run the full test (i.e. submitter + adaptor) is small then we can probably go with this approach. 

But if the adaptor takes significantly more time, then I think we should allow devs to be able to run only submitter tests.

karthikbekalp (commented):
Nothing blocking. Just few doubts and clarifications.

test/integ/helpers/submitter_test.py
joel-wong-aws: With a USD file, it's much easier to see what exactly is being rendered and compare it to the output. `.bip` files are binary and can't be inspected.

hatch.toml
joel-wong-aws: The submitter test takes about 30 seconds and the adaptor test takes about 70 seconds.

test/integ/helpers/submitter_test.py
joel-wong-aws: Currently, we don't. If the test runner wishes to do this, you can always do `git clean -xfd`.

test/integ/helpers/test_runners.py
joel-wong-aws: Yeah! Or maybe in https://github.com/aws-deadline/deadline-cloud-test-fixtures? Requesting to punt.

test/integ/helpers/submitter_test.py
karthikbekalp: I think it would be nice to have the runner cleanup test files if it were successful but I won't block on this.

hatch.toml
karthikbekalp: Hmm 70s is pretty quick, I guess devs could wait for a minute.

karthikbekalp (approved):

crowecawcaw (approved):

PR #193, opened 2025-04-09
docs: publish user guide to GitHub pages

crowecawcaw:
We have a user guide, but it's a file in a GitHub repo that's not easily discoverable and which visually looks like a developer-oriented doc.

Render the user guide as a static GitHub pages site using MkDocs.

There are many static site generators that read from markdown files, and MkDocs seemed like a common one for Python files. We can swap it out in the future with minimal effort.

There's a GitHub workflow which publishes the user guide as part of pushes to mainline.

See the user guide from my fork here: https://crowecawcaw.github.io/deadline-cloud-for-keyshot

End users have a nicer place to find install and usage info.

Published the user guide in my fork.

Yes

No

----

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -1,0 +1,33 @@
+name: Build and Deploy MkDocs
+
+on:
+  push:
+    branches:
+      - mainline
+    paths:
+      - 'docs/user_guide/**'
+      - 'mkdocs.yml'
+      - '.github/workflows/mkdocs.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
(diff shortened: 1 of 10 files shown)


baxeaz (approved):

.github/workflows/mkdocs.yml
```diff
@@ -0,0 +1,33 @@
+name: Build and Deploy MkDocs
```
joel-wong-aws, line 1: It would be great to put this in the [shared `.github` repository](https://github.com/aws-deadline/.github/tree/mainline/.github/workflows) for all integrations to use!

joel-wong-aws (approved):

PR #203, opened 2025-04-16
feat: add default locations for script installation on Mac and add UI hint

joel-wong-aws:
Since Mac does not support environment variables, the KeyShot scripts directory in the installer was always blank.

Added auto-detection for common KeyShot installation locations as well as some UI hints about where to look.

Easier installation for all users.

Windows:
`hatch build && hatch run lint && hatch run test && hatch run installer:build-installer --install-builder-path "C:\\Program Files\\InstallBuilder Professional 24.3.0" --platform windows --local-dev && hatch run test-installer`

Manually opened the installer and reviewed the text. Confirmed that the scripts directory was pre-filled.

image

Mac:
`hatch build && hatch run lint && hatch run test && hatch run installer:build-installer --install-builder-path "/Applications/Library Support/InstallBuilder Professional 24.3.0" --platform osx --local-dev && hatch run test-installer`

No, N/A

No

----

### DIFF
--- a/install_builder/deadline-cloud-for-keyshot.xml
+++ b/install_builder/deadline-cloud-for-keyshot.xml
@@ -71,7 +71,40 @@
                         <setInstallerVariable name="keyshot_scripts_directory" value="${env(KEYSHOT12)}\Scripts"/>
                     </actionList>
                     <elseActionList>
-                        <setInstallerVariable name="keyshot_scripts_directory" value=""/>
+                        <if>
+                            <conditionRuleList>
+                                <platformTest type="osx" />
+                            </conditionRuleList>
+                            <actionList>
+                                <if>
+                                    <conditionRuleList>
+                                        <!-- KeyShot 2024.2+ -->
+                                        <fileExists path="/Library/Application Support/KeyShot Studio/Scripts"/>
+                                    </conditionRuleList>
+                                    <actionList>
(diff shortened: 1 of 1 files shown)


install_builder/deadline-cloud-for-keyshot.xml
```diff
@@ -71,7 +71,40 @@
...
+                                        <!-- KeyShot 2024.2+ -->
+                                        <fileExists path="/Library/Application Support/KeyShot Studio/Scripts"/>
+                                    </conditionRuleList>
+                                    <actionList>
+                                        <setInstallerVariable name="keyshot_scripts_directory" value="/Library/Application Support/KeyShot Studio/Scripts"/>
+                                    </actionList>
+                                    <elseActionList>
+                                    <if>
+                                        <conditionRuleList>
+                                            <!-- KeyShot 2024.1 -->
```
epmog, line 90: If we support all these versions and we don't provide a user to specify which version to install for, what are your thoughts about installing for all the user's installed versions that we support/can see rather than prompting the user for 1 location? Are Windows locations are deterministic (ie. the env variables or what they resolve to)?

ie. If we see `/Library/Application Support/KeyShot Studio/Scripts` exists and `/Library/Application Support/KeyShot 12/Scripts` exists, we copy the script into both.

Issue #179, opened 2025-02-11
Add adaptor environment enter and exit timeouts


joel-wong-aws:
### Describe the bug

The KeyShot adaptor's environment enter and exits do not have a timeout specified in their [job template](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L195-L218).

### Expected Behaviour

The timeout for the environment enter should be a little more (ex. 1 minute) than the [server start timeout + KeyShot start timeout](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/18700054d5f396f5edd79857661b217992a428d7/src/deadline/keyshot_adaptor/KeyShotAdaptor/adaptor.py#L55-L58). The timeout for the environment exit should be a little more (ex. 1 minute) than the [server end timeout + KeyShot end timeout](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/18700054d5f396f5edd79857661b217992a428d7/src/deadline/keyshot_adaptor/KeyShotAdaptor/adaptor.py#L55-L58)

### Current Behaviour

If the environment enter or exit get stuck (and the first layer of timeouts doesn't work), then the worker agent does not enforce the timeout.

### Proposed Solution

Add a `timeout` to the environment enter and exit actions [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L195-L218), using the [OpenJD spec `timeout` field](https://github.com/OpenJobDescription/openjd-specifications/wiki/2023-09-Template-Schemas#5-action).

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): affects all Rendering OSes
2. Version of KeyShot: affects all versions of KeyShot
3. Version of this package: up to and including 0.3.1
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? N/A

Please share other details about your environment that you think might be relevant to reproducing the bug.

Example PR: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/173

crowecawcaw:
Recategorizing as an enhancement. It's definitely useful but isn't blocking current workflows.

joel-wong-aws:
[...]

PR #206, opened 2025-05-23
fix: add environment enter and exit timeouts

anika-suman-amazon:
Fixes: *https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/179*

The KeyShot adaptor's environment enter and exits do not have a timeout specified in their job template. If the environment enter or exit get stuck (and the first layer of timeouts doesn't work), then the worker agent does not enforce the timeout.

Added a timeout to the environment enter and exit actions, using the OpenJD spec timeout field. Added a unit test to test this change.

This change will prevent worker agents from getting stuck indefinitely if there are issues during KeyShot startup or shutdown.

This change was tested by creating an additional unit test (`test_construct_job_template_timeout_values`) and via `hatch run integ:test`.

No, as just the timeout values were added, along with a unit test and a change in the expected `template.json` file that adds in the timeout parameter.

No, it is not a breaking change.

----

### DIFF
--- a/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
+++ b/src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
@@ -19,6 +19,14 @@ SUBMISSION_MODE_KEY = "submission_mode"
 # Unique ID required to allow KeyShot to save selections for a dialog
 DEADLINE_CLOUD_DIALOG_ID = "e309ce79-3ee8-446a-8308-10d16dfcbb42"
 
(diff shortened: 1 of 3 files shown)


crowecawcaw:
Change looks good. Were you able to submit a job that used these timeouts and verify that it runs successfully?

anika-suman-amazon:
> Change looks good. Were you able to submit a job that used these timeouts and verify that it runs successfully?

Yes, I did test that, worked successfully!

crowecawcaw (approved):

joel-wong-aws (approved):

PR #212, opened 2025-06-06
fix: turn off viewport rendering to reduce CPU usage

### LINKED ISSUE #117: Bug: Viewport Rendering Starts when Keyshot launches
### Expected Behaviour

Keyshot starts rendering the viewport as soon as a scene is loaded. It seems to do this even when it's headless. By default Keyshot consumes as much CPU as possible. Viewport rendering should be turned off to save some cycles before "real" rendering begins. 

### Current Behaviour

As soon as the adaptor starts a keyshot client, the CPU spikes to 100%. 

### Reproduction Steps

Start keyshot in headless mode with any scene and a simple python script that infinitely sleeps.

### Code Snippet

Adding 
```python 
lux.pause()
```
to the top of https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py should fix the issue.

anika-suman-amazon:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/117

KeyShot initiates viewport rendering automatically upon scene loading, even in headless mode. By default, this process consumes maximum available CPU resources.

Adding lux.pause() in the init() method of keyshot_client.py to turn off viewport rendering.

Implementing lux.pause() disables viewport rendering, significantly reducing CPU utilization when the system is idle, allowing resources to be allocated more efficiently for actual rendering tasks when they begin.

I wrote a unit test for the change that passed, ran hatch integration tests, tested the changes in my own Conda environment, and then also took a look at both the Task Manager and Performance Monitor and visually compared how much % Processor Time python, keyshot_headless, and keyshot_openjd took up when running the integration tests with viewport rendering off, and then with it on. The differences demonstrate that CPU usage is indeed reduced.

The graphs from the Performance Monitor are attached here:
With viewport rendering turned off:
with pause

With viewport rendering turned on:
without pause

N/A

N/A

----

### DIFF
--- a/test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_client.py
+++ b/test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_client.py
@@ -16,3 +16,9 @@ def mock_get_keyshot_display_version():
 
 def test_keyshot_client_creation():
     KeyShotClient("127.0.01")
+
+
+def test_keyshot_client_pauses_viewport_rendering():
+    with mock.patch.object(lux, "pause") as pause_mock:
+        KeyShotClient("127.0.01")
+        pause_mock.assert_called_once()
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_client.py
@@ -46,6 +46,7 @@ class KeyShotClient(ClientInterface):
         super().__init__(server_path=server_path)
         major_version, minor_version = lux.getKeyShotDisplayVersion()
         print(f"KeyShotClient: KeyShot Version {major_version}.{minor_version}")
+        lux.pause()
         self.actions.update(KeyShotHandler().action_dict)
 
     def close(self, args: Optional[dict] = None) -> None:


crowecawcaw (approved):

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_client.py
```diff
@@ -16,3 +16,9 @@ def mock_get_keyshot_display_version():
 
 def test_keyshot_client_creation():
     KeyShotClient("127.0.01")
+
+
+def test_keyshot_client_pauses_viewport_rendering():
+    with mock.patch.object(lux, "pause") as pause_mock:
+        KeyShotClient("127.0.01")
+        pause_mock.assert_called_once()
```
joel-wong-aws, line 24: supernit: `pause_mock.assert_called_once()` could be outside of the `with` block to make it clearer that it is the assertion/testcase we are focusing on, and that it does not rely on being inside the context manager

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_client.py
```diff
@@ -16,3 +16,9 @@ def mock_get_keyshot_display_version():
 
 def test_keyshot_client_creation():
     KeyShotClient("127.0.01")
+
+
+def test_keyshot_client_pauses_viewport_rendering():
+    with mock.patch.object(lux, "pause") as pause_mock:
+        KeyShotClient("127.0.01")
```
joel-wong-aws, line 23: supernit: this was probably intended to be `KeyShotClient(127.0.0.1)`

joel-wong-aws (approved):

PR #214, opened 2025-06-12
fix!: Pass render data to adaptor and automate integration tests

### LINKED ISSUE #205: Bug: render settings not reflected in cloud render
### Describe the bug

As discussed in https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/180, some render settings in KeyShot such as maximum samples are not currently being applied to Deadline Cloud renders, unless the user first starts (but does not necessarily complete) a local render.

While this is actually a bug in KeyShot itself (defect ID "KEYS-32761"), we may be able to work around the issue by passing the render settings from the submitter to the adaptor.

### Expected Behaviour

Render settings set in KeyShot are applied to Deadline Cloud renders.

### Current Behaviour

Render settings are not applied.

### Reproduction Steps

Download the "Keyframe Animation" demo scene from [KeyShot](https://www.keyshot.com/keyshot-studio/scenes).

Adjust the maximum sample counts to a low value (e.g. 1) and submit the render via Deadline Cloud. 

The maximum samples setting will not be applied; the default maximum sample count of 1000 will still be in place.

### Proposed Solution

Approach 1 (preferred)
1. In the submitter, add data to the `init-data.yaml` [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L190) that includes `lux.getRenderOptions()` in the `init-data.yaml`.
2. In the adaptor, instead of using `lux.getRenderOptions()` [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/558c121e9cc6f2e5c41dad2883de32eb2e92033f/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py#L56), use the init data passed from the submitter.

Approach 2 (alternative if approach 1 doesn't work)
1. In the submitter, create an additional file containing the render options from `lux.getRenderOptions()`. Submit this file as part of the job
2. In the adaptor, check if the file exists, and if it does, read it and use it as render options.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): submitted from Windows 11 Enterprise 24H2, run on Deadline Cloud Service-Managed Fleets (Windows  Server 2022)
2. Version of KeyShot: 2025.1.1
3. Version of this package: 0.3.2
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? This can be reproduced on the 2025.04.21 submitter

anika-suman-amazon:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/205

The problem is that KeyShot render settings (i.e. maximum samples) aren't being applied to Deadline Cloud renders unless a local render is first initiated, which is a bug in KeyShot itself (defect ID "KEYS-32761"). The requirement is to implement a workaround by ensuring render settings are properly passed from the submitter to the adaptor, and to automate integration tests that verify these settings are correctly applied by comparing expected and actual output images.

In the submitter, I added the render options data to be included in the init-data.yaml. In the adaptor, instead of using lux.getRenderOptions(), I used the init data passed from the submitter. I also changed the value of progressive_max_samples from it’s original value (16) to 1000 for the integration test and changed the atol (tolerance level) for assert_all_images_close to 8 from 2, allowing me to automate the integration tests so the comparison check between the expected and actual output image passes.

This change enables automated verification that render settings are correctly applied in Deadline Cloud renders, ensuring consistent image quality without requiring users to start a local render first.

I wrote unit tests for the change for the handler, adaptor, and the submitter. I also ran the integration tests and set the progressive_max_samples setting within the integration test to see if they applied with my change. Lastly, I tested the changes in my Conda environment, changing the max samples in the Render Settings and determining if the output actually differed.

N/A

Yes, but I have made it so that it will stay backwards compatible by defaulting to the default KeyShot render settings if they are not received from the submitter.

----


crowecawcaw (dismissed):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -196,6 +200,7 @@ def construct_job_template(filename: str) -> dict:
                                         "scene_file: '{{Param.KeyShotFile}}'\n"
                                         "output_file_path: '{{Param.OutputFilePath}}'\n"
                                         "output_format: 'RENDER_OUTPUT_{{Param.OutputFormat}}'\n"
+                                        f"render_options: {yaml.dump(render_options_dict, default_flow_style=True, sort_keys=False).strip()}\n"
```
joel-wong-aws: KeyShot Python is missing a lot of standard libraries. Does it include `yaml`? To test this, we would need to uninstall (or temporarily move) the Python 3.11 so that KeyShot doesn't load it dynamically.

test/integ/helpers/test_runners.py
```diff
@@ -62,6 +63,21 @@ def run_keyshot_adaptor_test(
     with open(template_path) as f:
         template = yaml.safe_load(f)
 
+        for step in template["steps"]:
```
joel-wong-aws: nit: could this be simplified to something like 
```
with open(template_path, "rw") as f:
   new_content = f.read().replace("progressive_max_samples: 0", "progressive_max_samples: 1000")
   f.write(new_content)
```

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_handler.py
```diff
@@ -1,8 +1,36 @@
...
+        handler.render_kwargs = {
+            "render_options": {"engine_anti_aliasing": 1, "progressive_max_samples": 10},
+            "frame": 1,
+        }
+        handler.output_path = "test_%d.png"
+
+        handler.start_render({})
+
+        get_render_options_mock.assert_not_called()
+        render_image_mock.assert_called_once()
```
joel-wong-aws: nit: can we use `assert_called_once_with()` and verify that the expected render kwargs are passed in?

test/unit/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -57,7 +57,21 @@ def mock_lux_unpause():
...
+
+        def getDict(self):
+            return {"__VERSION": 5}
+
+    with mock.patch.object(submitter.lux, "getRenderOptions") as get_render_options_mock:
+        get_render_options_mock.return_value = MockRenderOptions()
+        yield get_render_options_mock
+
+
+def test_construct_job_template(mock_lux_get_render_options):
```
joel-wong-aws: If we're not actually using the `mock_lux_get_render_options` in these functions, can we instead just use `@pytest.fixture(autouse=True)` as the decorator on `def mock_lux_get_render_options`?

src/deadline/keyshot_adaptor/KeyShotAdaptor/adaptor.py
```diff
@@ -405,7 +405,12 @@ def on_run(self, run_data: dict) -> None:
             if name in run_data:
                 self._action_queue.enqueue_action(Action(name, {name: run_data[name]}))
 
-        self._action_queue.enqueue_action(Action("start_render", {"frame": run_data["frame"]}))
+        self._action_queue.enqueue_action(
+            Action(
+                "start_render",
+                {"frame": run_data["frame"], "render_options": self.init_data["render_options"]},
```
joel-wong-aws: If render_options isn't in `init_data`, will this throw an error? Can we make it handle that case?

joel-wong-aws (changes requested):

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -109,6 +110,9 @@ def construct_job_template(filename: str) -> dict:
     Constructs and returns a dict containing a valid job template for the KeyShot job.
     The return value is safe to convert/dump to JSON or YAML.
     """
+
+    render_options = re.sub(r'"([^"]+)":', r"\1:", json.dumps(lux.getRenderOptions().getDict()))
```
joel-wong-aws, line 115: Can we add a comment here about why we are doing this?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
anika-suman-amazon: I don't believe it includes yaml when I double checked, so I've changed the code to use only Python standard libraries! Thank you for the note. :)

test/integ/helpers/test_runners.py
anika-suman-amazon: Replaced it!

test/integ/cube/expected_bundle/template.json
```diff
@@ -92,7 +92,7 @@
...
+                                "data": "scene_file: '{{Param.KeyShotFile}}'\noutput_file_path: '{{Param.OutputFilePath}}'\noutput_format: 'RENDER_OUTPUT_{{Param.OutputFormat}}'\nrender_options: {add_to_queue: false, advanced_samples: 0, background_rendering: false, engine_anti_aliasing: 1, engine_caustics_quality: 0.0, engine_dof_quality: 3.0, engine_global_illumination: 0.0, engine_global_illumination_cache: true, engine_indirect_bounces: 1, engine_pixel_blur: 1.5, engine_ray_bounces: 6, engine_shadow_quality: 1.0, engine_sharp_shadows: true, engine_sharper_texture_filtering: false, engine_threads: 14, output_alpha_channel: false, output_ambient_occlusion_pass: false, output_caustics_pass: false, output_clown_pass: false, output_depth_pass: false, output_diffuse_pass: false, output_direct_lighting_pass: false, output_indirect_lighting_pass: false, output_labels_pass: false, output_normals_pass: false, output_raw_pass: false, output_reflection_pass: false, output_refraction_pass: false, output_render_layers: false, output_shadow_pass: false, progressive_max_samples: 0, progressive_max_time: 0.0, region: null, render_mode: 2, send_to_network: false, __VERSION: 5}\n"
```
joel-wong-aws, line 95: The render options here aren't valid JSON - the text isn't surrounded by quotes. How does this work? It seems like it would error without the quotes?

joel-wong-aws (changes requested):

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_handler.py
anika-suman-amazon: Changed it to asser_called_once_with()! I had to modify the assertion approach because the KeyShot API's renderImage function doesn't directly accept the render options dictionary as a parameter. Instead, it requires a RenderOptions object (opts) that's created internally by the handler. I'm not sure passing render_options as a parameter would work because the actual call signature uses different parameter names (path, opts, format) and the render options are encapsulated within the opts object rather than passed directly. But what I changed it to allows us to verify both that renderImage was called correctly and that the proper RenderOptions object was used.

test/unit/keyshot_submitter/test_keyshot_submitter.py
anika-suman-amazon: Done!

test/integ/cube/expected_bundle/template.json
anika-suman-amazon: Discussed that we should use yaml format rather than preserving the JSON

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
anika-suman-amazon: Done!

test/integ/cube/expected_bundle/template.json
joel-wong-aws: Right - `init-data.yaml` is indeed YAML (and [parsed as such](https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/blob/9648e754ca5883c010f403d3fb875c7146b9cc97/src/openjd/adaptor_runtime/_entrypoint.py#L502)), so the quotes aren't necessary here :)

joel-wong-aws (approved):

crowecawcaw (approved):

PR #217, opened 2025-06-19
fix!: Configure GPU rendering with host requirements via GUI

### LINKED ISSUE #151: Feature request: Configure GPU rendering with host requirements via GUI
### Describe the problem

There are two closely related problems to solve:
1. If a GPU render is submitted, a minimum of 1 GPU is not added to the host requirements (but it should be). This means the job could be assigned to the wrong fleet unless the user also selects a minimum of 1 GPU in the host requirements tab.
2. When submitting from a CPU-only workstation, Keyshot only allows the user to select a CPU-only render. It does not allow GPU renders to be configured. We should make this configurable in submissions to AWS Deadline Cloud.

### Proposed Solution

1. Add a job-specific setting "Render Engine" with options "CPU" or "GPU" to the submitter GUI by adding another option like [this](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/558c121e9cc6f2e5c41dad2883de32eb2e92033f/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L144-L165)
2. By default, if no previous option has been selected, use the `lux.getRenderEngine()` setting (`lux.RENDER_ENGINE_PRODUCT` and `lux.RENDER_ENGINE_INTERIOR` are CPU, `lux.RENDER_ENGINE_PRODUCT_GPU` and `lux.RENDER_ENGINE_INTERIOR_GPU` are GPU). Set that [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/558c121e9cc6f2e5c41dad2883de32eb2e92033f/src/deadline/keyshot_submitter/Submit%20to%20AWS%20Deadline%20Cloud.py#L513)
3. Pass the CPU or GPU option selected in the submitter GUI via the parameter values (should be automatic)
4. When receiving the parameter, if the scene has `lux.RENDER_ENGINE_PRODUCT` but the user has configured GPU, call `lux.setRenderEngine(lux.RENDER_ENGINE_PRODUCT_GPU)`. Apply similar logic to `lux.RENDER_ENGINE_INTERIOR` and vice versa. This logic would be in a new adaptor action `set_render_engine` added [here](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py#L14)
5. Investigate adding logic that if GPU is selected, add a requirement of minimum 1 GPU to the job's host requirements.
6. Add error handling logic - if the worker machine does not have a GPU, ensure a clear error message is raised.

Scripting reference: https://media.keyshot.com/scripting/doc/2024.3/lux.html

More information about lighting in Keyshot: https://www.keyshot.com/blog/using-lighting-presets-in-keyshot/

Note that CPU and GPU rendering can give different results, and neither is strictly better than the other: https://www.keyshot.com/blog/rendering-in-keyshot-an-analysis-of-cpu-and-gpu-rendering-performance

The proposed solution should not be a breaking change because if the parameter value is not passed to the adaptor, it should use the existing, default behaviour (use the settings from the scene).

### Example Use Cases

This will ensure that GPU renders are assigned to the correct workers and also allow users to submit a GPU render from a CPU-only workstation.

anika-suman-amazon:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/151

When users are working on a CPU-only workstation, KeyShot's interface only allows them to select CPU rendering options. This limitation prevents users from configuring GPU rendering jobs for AWS Deadline Cloud, even though the cloud workers might have GPUs available. We need to enhance the AWS Deadline Cloud submission process to allow users to select GPU rendering regardless of their local hardware capabilities. This way, users on CPU-only workstations can still take advantage of GPU rendering. Also, if a GPU render is submitted, a minimum of 1 GPU is not added to the host requirements, so the job could be assigned to the wrong fleet unless the user also selects a minimum of 1 GPU in the host requirements tab.

I added a "RenderEngine" parameter to the job template that allows users to select either CPU or GPU rendering, regardless of their local hardware. When GPU is selected, I automatically add a GPU requirement to the job template, ensuring the job is assigned to a worker with GPU capabilities. I also implemented error handling to provide clear messages if a job is assigned to a worker without the required GPU.

This change enables users on CPU-only workstations to submit GPU rendering jobs to AWS Deadline Cloud, allowing them to leverage cloud GPU resources for faster rendering. It also ensures jobs are properly routed to workers with the required hardware capabilities, preventing failed renders and improving overall workflow efficiency. Users will now have more flexibility in their rendering options regardless of their local hardware limitations.

I wrote and ran unit tests, I ran integration tests with hatch, and I ran it in my own Conda environment with both the GPU option and the CPU option. The result was that the GPU run was faster than the CPU. For a max sample value of 25, the CPU run took 8:37 minutes while the GPU run took 1:48 minutes for the same frame with the same worker, which goes in line with what we were expected because GPU renders should be faster if all other conditions remain the same. The submitter GUI change is also visible.

N/A

Yes

----

### DIFF
--- a/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
@@ -23,6 +23,9 @@
         "render_device": {
             "enum": ["CPU", "GPU"]
         },
+        "override_render_device": {
+            "type": "boolean"
+        },
         "render_options": {
             "type": "object"
         }
--- a/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
@@ -20,7 +20,7 @@
                 "RENDER_OUTPUT_PSD32"
             ]
         },
-        "render_engine": {
+        "render_device": {
             "enum": ["CPU", "GPU"]
         },
         "render_options": {
--- a/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
@@ -20,6 +20,9 @@
                 "RENDER_OUTPUT_PSD32"
             ]
         },
+        "render_engine": {
+            "enum": ["CPU", "GPU"]
+        },
         "render_options": {
             "type": "object"
         }
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
@@ -66,7 +66,7 @@ class KeyShotHandler:
         current_engine = lux.getRenderEngine()
         engine_name = engine_names[current_engine]
 
-        override_chosen = self.render_kwargs["override_render_device"]
+        override_chosen = self.render_kwargs.get("override_render_device", False)
         if override_chosen:
             print("Option to override the Keyshot render device was selected.")
             if self.render_kwargs["current_device"] == self.render_kwargs["render_device"]:
@@ -180,32 +180,31 @@ class KeyShotHandler:
         Raises:
             RuntimeError: If GPU rendering is requested but no compatible GPU is available
         """
-        override_render_device = self.render_kwargs["override_render_device"]
+        override_render_device = self.render_kwargs.get("override_render_device", False)
         self.original_render_device_param = data["render_device"]
 
         current_engine = lux.getRenderEngine()
         is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
         self.render_kwargs["current_device"] = "GPU" if is_gpu else "CPU"
+        self.render_kwargs["render_device"] = (
+            data["render_device"]
+            if override_render_device
+            else self.render_kwargs["current_device"]
+        )
 
-        if not override_render_device:
-            self.render_kwargs["render_device"] = "GPU" if is_gpu else "CPU"
-        else:
-            self.render_kwargs["render_device"] = data["render_device"]
-
-        if self.render_kwargs["render_device"] == "GPU":
-            if not is_gpu:
-                engine_map = {
-                    lux.RENDER_ENGINE_PRODUCT: lux.RENDER_ENGINE_PRODUCT_GPU,
-                    lux.RENDER_ENGINE_INTERIOR: lux.RENDER_ENGINE_INTERIOR_GPU,
-                }
-                if current_engine in engine_map:
-                    try:
-                        lux.setRenderEngine(engine_map[current_engine])
-                    except Exception:
-                        raise RuntimeError(
-                            "GPU rendering was requested but no compatible GPU is available on this worker. Please manually set min GPUs to 1 under 'Host requirements' in the KeyShot integrated submitter."
-                        )
-        elif override_render_device and self.render_kwargs["render_device"] == "CPU" and is_gpu:
+        if self.render_kwargs["render_device"] == "GPU" and not is_gpu:
+            engine_map = {
+                lux.RENDER_ENGINE_PRODUCT: lux.RENDER_ENGINE_PRODUCT_GPU,
+                lux.RENDER_ENGINE_INTERIOR: lux.RENDER_ENGINE_INTERIOR_GPU,
+            }
+            if current_engine in engine_map:
+                try:
+                    lux.setRenderEngine(engine_map[current_engine])
+                except Exception:
+                    raise RuntimeError(
+                        "GPU rendering was requested but no compatible GPU is available on this worker. Please manually set min GPUs to 1 under 'Host requirements' in the KeyShot integrated submitter."
+                    )
+        elif self.render_kwargs["render_device"] == "CPU" and is_gpu:
             engine_map = {
                 lux.RENDER_ENGINE_PRODUCT_GPU: lux.RENDER_ENGINE_PRODUCT,
                 lux.RENDER_ENGINE_INTERIOR_GPU: lux.RENDER_ENGINE_INTERIOR,
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
@@ -66,26 +66,30 @@ class KeyShotHandler:
         current_engine = lux.getRenderEngine()
         engine_name = engine_names[current_engine]
 
+        current_device = self.render_kwargs.get("current_device", get_current_render_device())
+        render_device = self.render_kwargs.get("render_device", current_device)
         override_chosen = self.render_kwargs.get("override_render_device", False)
+
         if override_chosen:
             print("Option to override the Keyshot render device was selected.")
-            if self.render_kwargs["current_device"] == self.render_kwargs["render_device"]:
+            if current_device == render_device:
                 print(
-                    f"KeyShot render device {self.render_kwargs['render_device']} is already selected. Chosen override option matches Keyshot setting.\n"
+                    f"KeyShot render device {render_device} is already selected. Chosen override option matches Keyshot setting.\n"
                 )
             else:
-                print(
-                    f"Overriding {self.render_kwargs['current_device']} with {self.render_kwargs['render_device']}...\n"
-                )
+                print(f"Overriding {current_device} with {render_device}...\n")
         else:
             print("Option to override the Keyshot render device was NOT selected.")
-            if self.render_kwargs["current_device"] != self.original_render_device_param:
+            if (
+                current_device != self.original_render_device_param
+                and self.original_render_device_param is not None
+            ):
                 print(
                     f"WARNING: RenderDevice value from the submitter ({self.original_render_device_param}) will be ignored since override is not selected.\n"
                 )
 
         print(f"Selected render engine: {engine_name}")
-        print(f"Selected render device: {self.render_kwargs['render_device']}")
+        print(f"Selected render device: {render_device}")
 
         print("Starting Render...")
         frame = self.render_kwargs["frame"]
@@ -182,17 +186,19 @@ class KeyShotHandler:
         """
         override_render_device = self.render_kwargs.get("override_render_device", False)
         self.original_render_device_param = data["render_device"]
-
         current_engine = lux.getRenderEngine()
-        is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
-        self.render_kwargs["current_device"] = "GPU" if is_gpu else "CPU"
+
+        self.render_kwargs["current_device"] = get_current_render_device()
         self.render_kwargs["render_device"] = (
             data["render_device"]
             if override_render_device
             else self.render_kwargs["current_device"]
         )
 
-        if self.render_kwargs["render_device"] == "GPU" and not is_gpu:
+        if (
+            self.render_kwargs["render_device"] == "GPU"
+            and self.render_kwargs["current_device"] == "CPU"
+        ):
             engine_map = {
                 lux.RENDER_ENGINE_PRODUCT: lux.RENDER_ENGINE_PRODUCT_GPU,
                 lux.RENDER_ENGINE_INTERIOR: lux.RENDER_ENGINE_INTERIOR_GPU,
@@ -204,10 +210,19 @@ class KeyShotHandler:
                     raise RuntimeError(
                         "GPU rendering was requested but no compatible GPU is available on this worker. Please manually set min GPUs to 1 under 'Host requirements' in the KeyShot integrated submitter."
                     )
-        elif self.render_kwargs["render_device"] == "CPU" and is_gpu:
+        elif (
+            self.render_kwargs["render_device"] == "CPU"
+            and self.render_kwargs["current_device"] == "GPU"
+        ):
             engine_map = {
                 lux.RENDER_ENGINE_PRODUCT_GPU: lux.RENDER_ENGINE_PRODUCT,
                 lux.RENDER_ENGINE_INTERIOR_GPU: lux.RENDER_ENGINE_INTERIOR,
             }
             if current_engine in engine_map:
                 lux.setRenderEngine(engine_map[current_engine])
+
+
+def get_current_render_device() -> str:
+    current_engine = lux.getRenderEngine()
+    is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
+    return "GPU" if is_gpu else "CPU"
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
@@ -32,6 +32,7 @@ class KeyShotHandler:
         self.render_kwargs = {}
         self.output_path = ""
         self.output_format_code = lux.RENDER_OUTPUT_PNG  # Default to PNG
+        self.original_render_device_param = None
 
     def set_output_file_path(self, data: dict) -> None:
         """
@@ -64,18 +65,27 @@ class KeyShotHandler:
 
         current_engine = lux.getRenderEngine()
         engine_name = engine_names[current_engine]
-        print(f"Selected render engine: {engine_name}")
-        selected_render_device = self.render_kwargs["render_device"]
-        print(f"Selected render device: {selected_render_device}")
-        override_chosen = self.render_kwargs["override_render_device"]
-        print(f"Override Keyshot render device chosen?: {override_chosen}\n")
 
-        if override_chosen and (
-            self.render_kwargs["current_device"] != self.render_kwargs["render_device"]
-        ):
-            print(
-                f"Overriding {self.render_kwargs['current_device']} with {self.render_kwargs['render_device']}...\n"
-            )
+        override_chosen = self.render_kwargs["override_render_device"]
+        if override_chosen:
+            print("Option to override the Keyshot render device was selected.")
+            if self.render_kwargs["current_device"] == self.render_kwargs["render_device"]:
+                print(
+                    f"KeyShot render device {self.render_kwargs['render_device']} is already selected. Chosen override option matches Keyshot setting.\n"
+                )
+            else:
+                print(
+                    f"Overriding {self.render_kwargs['current_device']} with {self.render_kwargs['render_device']}...\n"
+                )
+        else:
+            print("Option to override the Keyshot render device was NOT selected.")
+            if self.render_kwargs["current_device"] != self.original_render_device_param:
+                print(
+                    f"WARNING: RenderDevice value from the submitter ({self.original_render_device_param}) will be ignored since override is not selected.\n"
+                )
+
+        print(f"Selected render engine: {engine_name}")
+        print(f"Selected render device: {self.render_kwargs['render_device']}")
 
         print("Starting Render...")
         frame = self.render_kwargs["frame"]
@@ -158,7 +168,7 @@ class KeyShotHandler:
         Args:
             data (dict): The data given from the Adaptor. Keys expected: ['override_render_device']
         """
-        self.render_kwargs["override_render_device"] = data["override_render_device"]
+        self.render_kwargs["override_render_device"] = data.get("override_render_device", False)
 
     def set_render_device(self, data: dict) -> None:
         """
@@ -171,6 +181,7 @@ class KeyShotHandler:
             RuntimeError: If GPU rendering is requested but no compatible GPU is available
         """
         override_render_device = self.render_kwargs["override_render_device"]
+        self.original_render_device_param = data["render_device"]
 
         current_engine = lux.getRenderEngine()
         is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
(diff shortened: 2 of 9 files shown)


src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +126,38 @@ def set_render_options(self, data: dict) -> None:
         """
         if "render_options" in data:
             self.render_kwargs["render_options"] = data["render_options"]
+
+    def set_render_engine(self, data: dict) -> None:
```
crowecawcaw: Does the render engine appear in logs anywhere, maybe from KeyShot? If not, it could be useful info to print out at the end of this function.

src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
```diff
@@ -20,6 +20,9 @@
                 "RENDER_OUTPUT_PSD32"
             ]
         },
+        "render_engine": {
+            "enum": ["CPU", "GPU"]
+        },
```
crowecawcaw: Nit: there's a little bit of blurriness with what a render engine is in these changes. KeyShot's idea of a render engine is something like `RENDER_ENGINE_PRODUCT_GPU` while here we're using "render engine" to describe CPU vs GPU. For clarity, I think it'd be good to either use a different term here (e.g. "use_gpu") or pass in the full render engine (e.g. `RENDER_ENGINE_PRODUCT_GPU`) as the job parameter.

crowecawcaw (changes requested):
Great PR!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -584,6 +601,20 @@ def create_bundle(
     else:
         settings.parameter_values.append({"name": "KeyShotFile", "value": scene_file})
 
+    render_engine = None
+    if not any(param["name"] == "RenderEngine" for param in settings.parameter_values):
+        current_engine = lux.getRenderEngine()
+        is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
+        render_engine = "GPU" if is_gpu else "CPU"
```
joel-wong-aws: We use this logic twice in this file, can we move it into a function?

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_handler.py
```diff
@@ -4,6 +4,18 @@
...
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def setup_engine_constants():
+    """Set up the engine constants for all tests."""
+    lux.RENDER_ENGINE_PRODUCT = 0
+    lux.RENDER_ENGINE_INTERIOR = 1
+    lux.RENDER_ENGINE_PRODUCT_GPU = 2
+    lux.RENDER_ENGINE_INTERIOR_GPU = 3
```
joel-wong-aws: supernit: can we create constants for these outside of the pytest fixture so that it's easy to check later what `0`, `1`, `2`, and `3` mean? e.g.

```
RENDER_ENGINE_PRODUCT = 0
RENDER_ENGINE_INTERIOR = 1
RENDER_ENGINE_PRODUCT_GPU = 2
RENDER_ENGINE_INTERIOR_GPU = 3

@pytest.fixture(autouse=True)
def setup_engine_constants():
    """Set up the engine constants for all tests."""
    lux.RENDER_ENGINE_PRODUCT = RENDER_ENGINE_PRODUCT
    lux.RENDER_ENGINE_INTERIOR = RENDER_ENGINE_INTERIOR
    lux.RENDER_ENGINE_PRODUCT_GPU = RENDER_ENGINE_PRODUCT_GPU
    lux.RENDER_ENGINE_INTERIOR_GPU = RENDER_ENGINE_INTERIOR_GPU 
    yield
```

Or this might even work?
```
RENDER_ENGINE_PRODUCT = 0
RENDER_ENGINE_INTERIOR = 1
RENDER_ENGINE_PRODUCT_GPU = 2
RENDER_ENGINE_INTERIOR_GPU = 3

lux.RENDER_ENGINE_PRODUCT = RENDER_ENGINE_PRODUCT
lux.RENDER_ENGINE_INTERIOR = RENDER_ENGINE_INTERIOR
lux.RENDER_ENGINE_PRODUCT_GPU = RENDER_ENGINE_PRODUCT_GPU
lux.RENDER_ENGINE_INTERIOR_GPU = RENDER_ENGINE_INTERIOR_GPU 
```

Then, in our tests, we can do
```
set_render_engine_mock.assert_called_once_with(RENDER_ENGINE_PRODUCT_GPU)
```

instead of 
```
set_render_engine_mock.assert_called_once_with(2)
```

test/unit/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -397,3 +397,30 @@ def test_construct_job_template_timeout_values():
...
     assert actions["onEnter"]["timeout"] == submitter.KEYSHOT_ENVIRON_ENTER_TIMEOUT
     assert actions["onExit"]["timeout"] == submitter.KEYSHOT_ENVIRON_EXIT_TIMEOUT
+
+
+def test_construct_job_template_includes_render_engine_parameter():
+    with mock.patch.object(submitter.lux, "getRenderEngine") as get_render_engine_mock:
+        submitter.lux.RENDER_ENGINE_PRODUCT_GPU = 1
+        submitter.lux.RENDER_ENGINE_INTERIOR_GPU = 2
+
+        get_render_engine_mock.return_value = 0
```
joel-wong-aws: Similar note here about magic values and replacing them with constants.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -175,6 +179,18 @@ def construct_job_template(filename: str) -> dict:
                     "groupLabel": "KeyShot Settings",
                 },
             },
+            {
+                "name": "RenderEngine",
```
joel-wong-aws: Is this new setting sticky?

test/integ/cube/expected_bundle/template.json
```diff
@@ -58,6 +58,21 @@
                 "label": "Output Format (must match file extension)",
                 "groupLabel": "KeyShot Settings"
             }
+        },
+        {
+            "name": "RenderEngine",
```
joel-wong-aws: Is this new setting sticky?

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Added a log print statement for "Selected render device" to start_render! When GPU is chosen the logs automatically say GPU explicitly but they didn't explicitly have CPU in there before, so I added it for both.

src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
anika-suman-amazon: Changed all instances of "render_engine" to "render_device" to be more clear!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
anika-suman-amazon: Fixed!

test/unit/keyshot_adaptor/KeyShotClient/test_keyshot_handler.py
anika-suman-amazon: Fixed!

test/unit/keyshot_submitter/test_keyshot_submitter.py
anika-suman-amazon: Also fixed!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
anika-suman-amazon: Yep the last selected option stays when I relaunch the submitter!

test/integ/cube/expected_bundle/template.json
anika-suman-amazon: Yep the last selected option stays when I relaunch the submitter!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +133,37 @@ def set_render_options(self, data: dict) -> None:
...
+        """
+        Sets the render engine (CPU or GPU) based on the parameter value.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: ['render_device']
+
+        Raises:
+            RuntimeError: If GPU rendering is requested but no compatible GPU is available
+        """
+        # render_device = data["render_device"]
```
crowecawcaw: Nit: can be removed. There are a couple other commented out lines above too.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +56,10 @@ def start_render(self, data: dict) -> None:
         Raises:
             RuntimeError: .
         """
+
+        selected_render_device = self.render_kwargs["render_device"]
+        print(f"Selected render device: {selected_render_device}")
```
crowecawcaw: Nit: It might be slightly better to print out the render engine since it has both the device and the product vs interior engine info in it. It'd also be nice to print out the engine from lux.getRenderEngine() instead of from the job parameters so we can be more confident it was applied correctly.

crowecawcaw (changes requested):

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Makes sense, added the engine print as well! I kept the previous print as well because it makes sense to have it with the following print statement to see if the device is being overridden.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Done!

src/deadline/keyshot_adaptor/KeyShotAdaptor/schemas/init_data.schema.json
```diff
@@ -20,6 +20,12 @@
                 "RENDER_OUTPUT_PSD32"
             ]
         },
+        "render_device": {
+            "enum": ["CPU", "GPU"]
+        },
+        "override_render_device": {
+            "type": "boolean"
+        },
```
crowecawcaw, line 28: How would a user understand and use these two parameters together? Could we use just one? What does it mean to select CPU as the render device but not enable the override?

anika-suman-amazon: On the GUI, it shows up as "Override Keyshot render device" and the option is listed before the Render Device option. If they select override, the render device they select below will apply. Otherwise, it uses the KeyShot setting (CPU-mode vs GPU-mode) automatically. If they don't enable override the parameter for render device just doesn't do anything! This was because there wasn't a way for us to grey the setting out, so this was the best option for now.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +54,29 @@ def start_render(self, data: dict) -> None:
...
+        override_chosen = self.render_kwargs["override_render_device"]
+        print(f"Override Keyshot render device chosen?: {override_chosen}\n")
+
+        if override_chosen and (
+            self.render_kwargs["current_device"] != self.render_kwargs["render_device"]
+        ):
+            print(
+                f"Overriding {self.render_kwargs['current_device']} with {self.render_kwargs['render_device']}...\n"
+            )
+
```
joel-wong-aws, line 93: The printout here is a bit confusing. I think the logic we want is:

1. If override is selected:
    * Print that override is selected (first)
    * If the override value == keyshot value:
        * Print that the override matches the KeyShot setting and that we are proceeding
    * If the override value != keyshot setting 
        * Print the override value like you've done here
2. If override is not selected
    * Print that render device is not overridden (first)
    * If the override value != keyshot value:
        * Print a warning that the render device value will be ignored since the override is deselected
    * If the override value == keyshot value, no need to print anything else

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +150,54 @@ def set_render_options(self, data: dict) -> None:
...
+        if self.render_kwargs["render_device"] == "GPU":
+            try:
+                if not is_gpu:
+                    engine_map = {
+                        lux.RENDER_ENGINE_PRODUCT: lux.RENDER_ENGINE_PRODUCT_GPU,
+                        lux.RENDER_ENGINE_INTERIOR: lux.RENDER_ENGINE_INTERIOR_GPU,
+                    }
+                    if current_engine in engine_map:
+                        lux.setRenderEngine(engine_map[current_engine])
+            except Exception:
```
joel-wong-aws: nit: we only expect this exception to occur during the `lux.setRenderEngine()` call, right? Can we minimize the scope of the try-except to wrap around only that line instead to be more precise and avoid swallowing unexpected errors?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -584,6 +612,39 @@ def create_bundle(
...
 
+    override_enabled = False
+    for param in settings.parameter_values:
+        if param["name"] == "OverrideRenderDevice":
+            override_enabled = param["value"] == "TRUE"
+            break
+
+    if not any(param["name"] == "OverrideRenderDevice" for param in settings.parameter_values):
+        settings.parameter_values.append({"name": "OverrideRenderDevice", "value": "FALSE"})
+        override_enabled = False
```
joel-wong-aws: nit: could this logic be combined with the previous `for` statement?

test/integ/cube/expected_bundle/parameter_values.json
```diff
@@ -12,6 +12,14 @@
             "name": "OutputFormat",
             "value": "PNG"
         },
+        {
+            "name": "RenderDevice",
+            "value": "CPU"
+        },
+        {
+            "name": "OverrideRenderDevice",
+            "value": "TRUE"
```
joel-wong-aws: Isn't this FALSE by default?

joel-wong-aws:
Looking good, just commented about some logging improvements that will help customers debug what's going on with their render device choices :)

test/integ/cube/expected_bundle/parameter_values.json
anika-suman-amazon: Fixed!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
anika-suman-amazon: I couldn't quite combine it but I did simplify it in my new revision so it doesn't repeat some of the logic!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Done!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Added :)

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +55,38 @@ def start_render(self, data: dict) -> None:
...
+            0: "RENDER_ENGINE_PRODUCT",
+            1: "RENDER_ENGINE_INTERIOR",
+            3: "RENDER_ENGINE_PRODUCT_GPU",
+            4: "RENDER_ENGINE_INTERIOR_GPU",
+        }
+
+        current_engine = lux.getRenderEngine()
+        engine_name = engine_names[current_engine]
+
+        override_chosen = self.render_kwargs["override_render_device"]
```
joel-wong-aws: If the `override_render_device` is not passed in the init data, this will error (right)? We need `override_render_device = self.render_kwargs.get("override_render_device", False)` here

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +160,55 @@ def set_render_options(self, data: dict) -> None:
...
+        """
+        Sets the render engine (CPU or GPU) based on the parameter value.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: ['render_device']
+
+        Raises:
+            RuntimeError: If GPU rendering is requested but no compatible GPU is available
+        """
+        override_render_device = self.render_kwargs["override_render_device"]
```
joel-wong-aws: If the `override_render_device` is not passed in the init data, this will error (right)? We need `override_render_device = self.render_kwargs.get("override_render_device", False)` here

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +160,55 @@ def set_render_options(self, data: dict) -> None:
...
+        is_gpu = current_engine in [lux.RENDER_ENGINE_PRODUCT_GPU, lux.RENDER_ENGINE_INTERIOR_GPU]
+        self.render_kwargs["current_device"] = "GPU" if is_gpu else "CPU"
+
+        if not override_render_device:
+            self.render_kwargs["render_device"] = "GPU" if is_gpu else "CPU"
+        else:
+            self.render_kwargs["render_device"] = data["render_device"]
+
+        if self.render_kwargs["render_device"] == "GPU":
+            if not is_gpu:
```
joel-wong-aws: supernit: this `if` statement can be combined with the previous `if` statement, right?

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -125,3 +160,55 @@ def set_render_options(self, data: dict) -> None:
...
+                    lux.RENDER_ENGINE_INTERIOR: lux.RENDER_ENGINE_INTERIOR_GPU,
+                }
+                if current_engine in engine_map:
+                    try:
+                        lux.setRenderEngine(engine_map[current_engine])
+                    except Exception:
+                        raise RuntimeError(
+                            "GPU rendering was requested but no compatible GPU is available on this worker. Please manually set min GPUs to 1 under 'Host requirements' in the KeyShot integrated submitter."
+                        )
+        elif override_render_device and self.render_kwargs["render_device"] == "CPU" and is_gpu:
```
joel-wong-aws: supernit: the `override_render_device` condition is redundant here, right?

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: I believe it is already taken care of when we set render_kwargs for override_render_device in set_override_render_device...we do self.render_kwargs["override_render_device"] = data.get("override_render_device", False). But I guess it doesn't hurt to change that too, I'll update it!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Yep! Fixed!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Also fixed!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Fixed!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
joel-wong-aws: Technically, `override_render_device` doesn't have to be passed in the init-data according to the `init_data.schema.json`, so `set_override_render_device` may not be called at all!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +55,38 @@ def start_render(self, data: dict) -> None:
...
+                )
+        else:
+            print("Option to override the Keyshot render device was NOT selected.")
+            if self.render_kwargs["current_device"] != self.original_render_device_param:
+                print(
+                    f"WARNING: RenderDevice value from the submitter ({self.original_render_device_param}) will be ignored since override is not selected.\n"
+                )
+
+        print(f"Selected render engine: {engine_name}")
+        print(f"Selected render device: {self.render_kwargs['render_device']}")
```
joel-wong-aws: All the `self.render_kwargs['render_device']` and `self.render_kwargs["current_device"]` in this function may also error if the job is submitted using a previous submitter without `render_device` in the init-data, can we handle those cases too? The only guaranteed argument is `self.render_kwargs["frame"]`

anika-suman-amazon: Done!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Oh right, understood!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +55,39 @@ def start_render(self, data: dict) -> None:
...
+            1: "RENDER_ENGINE_INTERIOR",
+            3: "RENDER_ENGINE_PRODUCT_GPU",
+            4: "RENDER_ENGINE_INTERIOR_GPU",
+        }
+
+        current_engine = lux.getRenderEngine()
+        engine_name = engine_names[current_engine]
+
+        current_device = self.render_kwargs.get("current_device", get_current_render_device())
+        render_device = self.render_kwargs.get("render_device", "CPU")
```
joel-wong-aws: The default here should be `current_device`, i.e. `render_device = self.render_kwargs.get("render_device", current_device)` right? Otherwise we'll get unexpected output later in the function.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +55,39 @@ def start_render(self, data: dict) -> None:
...
+            print("Option to override the Keyshot render device was selected.")
+            if current_device == render_device:
+                print(
+                    f"KeyShot render device {render_device} is already selected. Chosen override option matches Keyshot setting.\n"
+                )
+            else:
+                print(f"Overriding {current_device} with {render_device}...\n")
+        else:
+            print("Option to override the Keyshot render device was NOT selected.")
+            if current_device != self.original_render_device_param:
```
joel-wong-aws: `if current_device != self.original_render_device_param and self.original_render_device_param is not None`:

anika-suman-amazon: Updated!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Fixed!

joel-wong-aws (approved):

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -52,6 +55,42 @@ def start_render(self, data: dict) -> None:
...
+
+        engine_names = {
+            0: "RENDER_ENGINE_PRODUCT",
+            1: "RENDER_ENGINE_INTERIOR",
+            3: "RENDER_ENGINE_PRODUCT_GPU",
+            4: "RENDER_ENGINE_INTERIOR_GPU",
+        }
+
+        current_engine = lux.getRenderEngine()
+        engine_name = engine_names[current_engine]
```
crowecawcaw, line 67: If KeyShot updates its enums to add a new render engine, this call will start to fail. We could make it more resilient by adding a fallback option, maybe the int value. Or even better if we can pull this from KeyShot's SDK somehow. Not blocking though!

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -584,6 +612,38 @@ def create_bundle(
     else:
         settings.parameter_values.append({"name": "KeyShotFile", "value": scene_file})
 
+    override_enabled = False
+    for param in settings.parameter_values:
```
crowecawcaw, line 616: Nit: we do this loop three time. Code might be simpler if we combined them.

crowecawcaw (approved):

PR #221, opened 2025-07-01
fix: fixed override_param_found submitter bug

anika-suman-amazon:
A bug in the submitter that didn't set a value for override_param_found without sticky settings.

Just added one line setting override_param_found = False.

Fixes a bug in the submitter that didn't set a value for override_param_found without sticky settings.

Tested with unit tests, integration tests, and with a new conda package build.

N/A

No

----


joel-wong-aws (approved):

crowecawcaw (approved):
It'd be great to go back and add or expand tests so that they cover this change. For this change to be covered, we want tests to fail before without this change and pass with it.

PR #228, opened 2025-07-09
fix: Refined KeyShot version mismatch error message

anika-suman-amazon:
KeyShot error message was misleading.

Changed "WARNING: Version mismatch detected but continuing: This scene was saved using a newer version [14.0]. Opening it in an older version will cause some information to be lost. We recommend updating to version [14.0] or later." to "WARNING: Version mismatch detected but continuing: This scene was saved using a newer version [14.0]. Opening it in an older version may cause some information to be lost."

Prevent misleading customers.

Unit tests, integration tests, and testing with my Conda environment.

N/A

No

----

### DIFF
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
@@ -110,7 +110,9 @@ class KeyShotHandler:
         except Exception as e:
             error_message = str(e)
             if "This scene was saved using a newer version" in error_message:
-                print(f"WARNING: Version mismatch detected but continuing: {error_message}")
+                updated_error_message = error_message.split("We recommend")[0].strip()
+                updated_error_message = updated_error_message.replace("will", "may")
+                print(f"WARNING: Version mismatch detected but continuing: {updated_error_message}")
             else:
                 raise
         print(f"Finished Rendering {output_path}")
--- a/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
+++ b/src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
@@ -110,8 +110,7 @@ class KeyShotHandler:
         except Exception as e:
             error_message = str(e)
             if "This scene was saved using a newer version" in error_message:
-                updated_error_message = error_message.split("We recommend")[0].strip()
-                updated_error_message = updated_error_message.replace("will", "may")
+                updated_error_message = "This scene was saved using a newer version. Opening it in an older version may cause some information to be lost."
                 print(f"WARNING: Version mismatch detected but continuing: {updated_error_message}")
             else:
                 raise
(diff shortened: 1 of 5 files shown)


src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -110,7 +110,9 @@ def start_render(self, data: dict) -> None:
         except Exception as e:
             error_message = str(e)
             if "This scene was saved using a newer version" in error_message:
-                print(f"WARNING: Version mismatch detected but continuing: {error_message}")
+                updated_error_message = error_message.split("We recommend")[0].strip()
+                updated_error_message = updated_error_message.replace("will", "may")
```
karthikbekalp: Its hard to say what message is being replaced here. Can we replace the full error message with what we want to convey to the customer by extracting the dynamic parts from the message?

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -110,7 +110,9 @@ def start_render(self, data: dict) -> None:
...
+                print(f"WARNING: Version mismatch detected but continuing: {updated_error_message}")
```
karthikbekalp, line 114: I wonder if we should even change the error message from the DCC. There's probably reason why they added it and recommend updating no?

anika-suman-amazon: Previously the KeyShot error was leading to the job failing but since the render was completing we decided to just ignore the message as it looks like a false alarm! For this PR we just wanted to remove the "We recommend updating to version [14.0] or later." part of that error message because it can be misleading for customers.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
anika-suman-amazon: Done!

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
karthikbekalp: Makes sense. Thanks for your clarification.

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
karthikbekalp: Thanks, this is much clearer.

karthikbekalp (approved):

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -110,7 +110,8 @@ def start_render(self, data: dict) -> None:
         except Exception as e:
             error_message = str(e)
             if "This scene was saved using a newer version" in error_message:
-                print(f"WARNING: Version mismatch detected but continuing: {error_message}")
+                updated_error_message = "This scene was saved using a newer version. Opening it in an older version may cause some information to be lost."
+                print(f"WARNING: Version mismatch detected but continuing: {updated_error_message}")
```
karthikbekalp, line 114: nit: This could've been a single line instead of using `updated_error_message` variable. But no issues.

joel-wong-aws (approved):

Issue #191, opened 2025-03-21
Submit whole render queue at once


farrahandpearce:
### Describe the problem

When you want to submit multiple scenes this is fundimental to how Keyshot performs. It's inpractical to have to sumit for every viewpoint / material option / lighting set up. 

### Proposed Solution

Be able to fill up a render queue with a variety of jobs, then submit the whole thing as a combined job.

### Example Use Cases

N/A

joel-wong-aws:
@farrahandpearce Thank you for opening this issue!

I've added this feature request to our backlog.

A few questions about your use case:

1. Do you usually use KeyShot's functionality to "Group into a single render job" when adding different viewpoints / materials / lighting options or do you uncheck that option?

2. Do you typically have KeyShot jobs from multiple projects in your render queue when you want to submit it to Deadline Cloud?
 
3. When you say "combined job" is your goal to have a single workflow/script which submits everything in the render queue? Or are you specifically looking to submit the entire render queue as a single Deadline Cloud job (with multiple steps, one for each KeyShot job in the render queue)?
    * We might have some technical constraints around a single job vs multiple jobs, but if given the option, do you have a preference?
    * For more info about the differences between Deadline Cloud jobs and steps, see [these docs](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/deadline-cloud-jobs.html)

farrahandpearce:
Do you usually use KeyShot's functionality to "Group into a single render job" when adding different viewpoints / materials / lighting options or do you uncheck that option? - _Better not to group the jobs, only as then you cannot see how long each is taking afterwards, so that settings can be modified and rerendered etc._

Do you typically have KeyShot jobs from multiple projects in your render queue when you want to submit it to Deadline Cloud? _Sometimes, but I don't think this is a dealbreaker as it would be acceptable to submit one queue per file._

[...]

PR #233, opened 2025-07-22
feat: Add a render queue submission feature for different camera options

anika-suman-amazon:
Fixes (partially): https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/191

Right now, the Deadline Cloud integration with KeyShot does not allow the user to submit multiple scenes at once, with different combinations of cameras/viewpoints, colorways, model sets, lighting options, etc., reducing efficiency as they instead have to submit one at a time. This update allows the user to render multiple cameras at once. The other settings are more difficult to access with the KeyShot API available to us, so we would need to contact KeyShot to proceed with adding the other options to make this feature even more robust.

I added a new dialog box to allow the user to select which cameras out of all available camera views the user wants to render. Then, I made it so each camera option added a new step to the job template so that all of them could be rendered with one job. 

Increases efficiency and ease for customers, allowing them to render multiple camera viewpoints at once.

I tested against the unit tests, adjusting where needed, and I also tested with the integration tests and by creating a Conda package with my changes.

N/A

Yes, changes have been made to the adaptor.

----


src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -102,6 +103,10 @@ def start_render(self, data: dict) -> None:
         opts.setAddToQueue(False)
         lux.setAnimationFrame(frame)
         output_path = self.output_path.replace("%d", str(frame))
+
+        output_dir = os.path.dirname(output_path)
+        print(f"Requested output directory: {output_dir}")
```
joel-wong-aws, line 108: left over debug statement?

src/deadline/keyshot_adaptor/KeyShotClient/keyshot_handler.py
```diff
@@ -230,6 +235,25 @@ def set_render_device(self, data: dict) -> None:
...
+        """
+        Sets the camera to use for rendering.
+
+        Args:
+            data (dict): The data given from the Adaptor. Keys expected: ['camera']
+        """
+        camera = data.get("camera", lux.getCamera())
+
+        print(f"Setting camera to: {camera}")
+        print(f"lux.getCamera() result: {lux.getCamera()}")
```
joel-wong-aws: Leftover debug statement?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -292,6 +307,32 @@ def construct_job_template(filename: str) -> dict:
...
+                "type": "PATH",
+                "objectType": "FILE",
+                "dataFlow": "OUT",
+                "userInterface": {
+                    "control": "HIDDEN",
+                    "label": f"Output Path for {label}",
+                    "groupLabel": "KeyShot Settings",
+                },
+                "description": f"The render output path {label}.",
+                "default": "{{Param.OutputFilePath}}",
```
joel-wong-aws: This is a reference to a different output job parameter, which isn't valid. We shouldn't include a default value if there is no reasonable default.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
+                    is_selected,
+                )
+            )
+
+    if not render_dialog_items:
+        return file_selections
+
+    render_selections = lux.getInputDialog(
+        title="Select Cameras to Render",
+        values=render_dialog_items,
```
joel-wong-aws, line 534: Why did we choose to put this in a separate dialog, rather than the same dialog? (I recall there was a reason for it, but if we can document that in code, it would be helpful for future developers)

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
-        values=dialog_items,
+        values=file_dialog_items,
         id=DEADLINE_CLOUD_DIALOG_ID,
     )
 
+    cameras = lux.getCameras()
+    active_camera = lux.getCamera()
+
+    if len(cameras) == 1:
+        file_selections["selected_cameras"] = [cameras[0]]
```
joel-wong-aws: Can we make `"selected_cameras"` a shared constant?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
+                (
+                    f"camera_{i}",
+                    lux.DIALOG_CHECK,
+                    label,
+                    is_selected,
+                )
+            )
+
+    if not render_dialog_items:
+        return file_selections
```
joel-wong-aws: Will this conditional ever be true? If not, can we remove it?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
+    cameras = lux.getCameras()
+    active_camera = lux.getCamera()
+
+    if len(cameras) == 1:
+        file_selections["selected_cameras"] = [cameras[0]]
+        return file_selections
+
+    render_dialog_items: list[tuple[Any, ...]] = []
+
+    if cameras:
```
joel-wong-aws: Are there any scenarios where this conditional is False? If not, we can remove the `if` statement

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
+        values=file_dialog_items,
         id=DEADLINE_CLOUD_DIALOG_ID,
     )
 
+    cameras = lux.getCameras()
+    active_camera = lux.getCamera()
+
+    if len(cameras) == 1:
+        file_selections["selected_cameras"] = [cameras[0]]
+        return file_selections
```
joel-wong-aws: One way we can simplify this block (and make it a little safer) is to instead do
```
file_selections["selected_cameras"] = [cameras[0]]
if len(cameras) == 1:
    return file_selections
```
That way, even if there is an issue later, we can always return `file_selections` and know it will have a reasonable result.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -425,12 +467,63 @@ def options_dialog(show_gui=True) -> dict[str, Any]:
...
+    selections = file_selections
+
+    selected_cameras = []
+    for i in range(len(cameras)):
+        key = f"camera_{i}"
+        if key in render_selections and render_selections[key]:
+            selected_cameras.append(cameras[i])
+
+    if not selected_cameras and cameras:
+        selected_cameras = ["last_active"]
```
joel-wong-aws: Should we raise an error if the user does not select any cameras instead?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -653,17 +746,77 @@ def create_bundle(
             "value": f"keyshot={major_version}.* keyshot-openjd=0.4.*",
         }
     )
+    # settings.parameter_values.append({"name": "CondaChannels", "value": "deadline-cloud"})
```
joel-wong-aws: Leftover debug statement

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -653,17 +746,77 @@ def create_bundle(
...
     settings.parameter_values.append({"name": "CondaChannels", "value": "deadline-cloud"})
 
     job_template = construct_job_template(scene_name)
+
+    selected_cameras = dialog_selections.get("selected_cameras", [])
+
+    if selected_cameras:
```
joel-wong-aws: Is there any way this can be false? If not, can remove this conditional statement?

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -653,17 +746,77 @@ def create_bundle(
...
+        steps = []
+
+        base_output_path = ""
+        for param in settings.parameter_values:
+            if param["name"] == "OutputFilePath":
+                base_output_path = param["value"]
+                break
+
+        originalStep = copy.deepcopy(job_template["steps"][0])
```
joel-wong-aws: nit: `originalStep` -> `original_step` (Python convention is to use `snake_case` rather than `camelCase`). Same for `updatedStep`

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -653,17 +746,77 @@ def create_bundle(
...
+        original_init_data = originalStep["stepEnvironments"][0]["script"]["embeddedFiles"][0][
+            "data"
+        ]
+
+        for camera in selected_cameras:
+            updatedStep = copy.deepcopy(originalStep)
+            updatedStep["name"] = f"Render_{camera}"
+
+            camera_safe_name = (
+                camera.replace(" ", "_").replace("/", "_").replace("&", "and").replace("-", "_")
```
joel-wong-aws: We do this in multiple places, can we use a function for it instead?

test/integ/cube/expected_bundle/template.json
```diff
@@ -88,6 +99,19 @@
...
+            "type": "PATH",
+            "objectType": "FILE",
+            "dataFlow": "OUT",
+            "userInterface": {
+                "control": "HIDDEN",
+                "label": "Output Path for Current Active Camera",
+                "groupLabel": "KeyShot Settings"
+            },
+            "description": "The render output path Current Active Camera.",
+            "default": "{{Param.OutputFilePath}}"
```
joel-wong-aws: The "default value" here should be the actual output path.

src/deadline/keyshot_submitter/Submit to AWS Deadline Cloud.py
```diff
@@ -653,17 +746,77 @@ def create_bundle(
...
+
+            param_found = False
+            for param in settings.parameter_values:
+                if param["name"] == param_name:
+                    param["value"] = camera_output_path
+                    param_found = True
+                    break
+
+            if not param_found:
+                settings.parameter_values.append({"name": param_name, "value": camera_output_path})
```
joel-wong-aws: It could also be helpful to save the `camera_output_path` as the default value for the `param_name` job parameters here.

test/integ/cube/expected_bundle/template.json
```diff
@@ -122,7 +146,7 @@
...
...utFilePath}}'\noutput_format: 'RENDER_OUTPUT_{{Param.OutputFormat}}'\noverride_render_device: {{Param.OverrideRenderDevice}}\nrender_device: '{{Param.RenderDevice}}'\ncamera: '{{Param.Camera}}'\nrender_options: {add_to_queue: false, advanced_samples: 0, background_rendering: false, engine_anti_aliasing: 1, engine_caustics_quality: 0.0, engine_dof_quality: 3.0, engine_global_illumination: 0.0, engine_global_illumination_cache: true, engine_indirect_bounces: 1, engine_pixel_blur: 1.5, engine_ray_bounces: 6, engine_shadow_quality: 1.0, engine_sharp_shadows: true, engine_sharper_texture_filtering: false, engine_threads: 14, output_alpha_channel: false, output_ambient_occlusion_pass: false, output_caustics_pass: false, output_clown_pass: false, output_depth_pass: false, output_diffuse_pass: false, output_direct_lighting_pass: false, output_indirect_lighting_pass: false, output_labels_pass: false, output_normals_pass: false, output_raw_pass: false, output_reflection_pass: false, output_refraction_pass: false, output_render_layers: false, output_shadow_pass: false, progressive_max_samples: 0, progressive_max_time: 0.0, region: null, render_mode: 2, send_to_network: false, __VERSION: 5}\n"
```
joel-wong-aws: Out of scope for this PR, but we should add an integration test where there are multiple cameras to test the new flow.

test/unit/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -76,12 +76,29 @@ def getDict(self):
...
+
 def test_construct_job_template():
     filename = "test_filename"
 
-    job_template = submitter.construct_job_template(filename)
+    with mock.patch.object(submitter.lux, "getCameras", return_value=[]):
+        job_template = submitter.construct_job_template(filename)
 
-    assert job_template["name"] == filename
+        assert job_template["name"] == filename
```
joel-wong-aws: supernit: can we leave the assert unindented? The `job_template` also exists outside of the mock and the output could theoretically be affected by the mock (albeit not in this case), so I think it's better to leave it outside the mock.

Similar note for the asserts below.

test/unit/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -395,17 +412,20 @@ def test_get_ksp_bundle_files():
 
 
 def test_construct_job_template_timeout_values():
+    with mock.patch.object(submitter.lux, "getCameras", return_value=[]):
```
joel-wong-aws: Is an empty list for `getCameras` possible? The test data should replicate some realistic data

Also, we should probably add this as an autoused fixture `def mock_lux_get_cameras():`

test/unit/keyshot_submitter/test_keyshot_submitter.py
```diff
@@ -395,17 +412,20 @@ def test_get_ksp_bundle_files():
...
 
 
 def test_construct_job_template_includes_render_device_parameter():
-    with mock.patch.object(submitter.lux, "getRenderEngine") as get_render_device_mock:
+    with (
+        mock.patch.object(submitter.lux, "getRenderEngine") as get_render_device_mock,
+        mock.patch.object(submitter.lux, "getCameras", return_value=[]),
+    ):
         submitter.lux.RENDER_ENGINE_PRODUCT_GPU = RENDER_ENGINE_PRODUCT_GPU
         submitter.lux.RENDER_ENGINE_INTERIOR_GPU = RENDER_ENGINE_INTERIOR_GPU
```
joel-wong-aws, line 435: We either need an integration or unit test for the new flow with multiple cameras :)

joel-wong-aws (changes requested):

joel-wong-aws:
This is looking really good! I've added a bunch of comments but they are mostly around better handling edge cases and remove some redundant logic. This is a great step towards addressing https://github.com/aws-deadline/deadline-cloud-for-keyshot/issues/191 :)

PR #238, opened 2025-08-14
feat!: Support KeyShot 2025, remove dependency on adapter package

crowecawcaw:
Two problems:
- the current KeyShot adapter has dependencies which need to be loaded from a system Python install which is fickly and risk. This dependency loading behavior is due to KeyShot's limited standard libraries in its bundled Python and its unconventional approach of finding and loading dependencies from system Python installs.
- On Deadline SMF, the current system Python install is 3.11 which is not compatible with KeyShot 2024.3 and 2025, so the required dependencies cannot be loaded

Beyond these immediate problems, developing, updating, releasing the adapter is tedious because it needs to align with the submitter but is released and distributed as a separate Python application.

- ### What was the solution? (How)
- Embed the adapter scripts directly in the job template instead of distributing them as a separate Python application
- Update the adapter to remove the OJD adapter runtime dependencies
- Communicate between processes through sockets instead of named pipes to avoid a pywin32 dependency

For a good overview of the architecture, read the rendered README that's included in this CR: https://github.com/crowecawcaw/deadline-cloud-for-keyshot/blob/filestream/src/README.md

- supports KeyShot 2025 on Deadline SMF since the adapter no longer depends on the Python system install matching KeyShot's internal Python version
- eliminates version matching requirements between the submitter and adapter since they're always bundled together
- simplifies development, maintenance, and releases since we no longer have a separate adapter Python package

- Submitted lots of jobs manually
- Added unit tests
- Ran integ tests

Yes, see the new ./src/README.md file

It doesn't break existing workflows, but it is a major change to the architecture.

### TODO on this draft
- test GPU overrides (it's implemented but I need to manually test)
- support render settings (it's passed in as an arg but not applied
----


pyproject.toml
```diff
@@ -11,14 +11,12 @@ authors = [
 dynamic = ["version"]
 readme = "README.md"
 license = "Apache-2.0"
-requires-python = ">=3.9"
+requires-python = ">=3.1"
```
diane-hj: should this be 3.11? 👀

src/README.md
```diff
@@ -0,0 +1,214 @@
...
+        ADP --> KSW
+    end
+    
+    SUB -->|Submit job| DC
+    DC -->|Schedule tasks| ADP
+```
+
+## Submitter
+
+The submitter is a Python script that users can invoke from KeyShot's scripting console. The script is run by KeyShot's embedded Python interpreter which has access to the `lux` API. The submitter uses the `lux` API to extract scene information, manage file dependencies, and generate an OpenJobDescription job bundle. One the bundle is generated, the script invokes the Deadline Client job bundle submitter (`deadline bundle gui-submit`) which presents the job submission dialog.
```
diane-hj: nit: typo in `one the bundle is generated` --> `once the bundle is generated`

src/keyshot_command_handler.py
```diff
@@ -0,0 +1,158 @@
...
+                updated_error_message = "This scene was saved using a newer version. Opening it in an older version may cause some information to be lost."
+                print(f"WARNING: Version mismatch detected but continuing: {updated_error_message}")
+            else:
+                raise
+
+        print(f"Finished Rendering {output_path}")
+        print('{"status": "success", "frame": ' + str(frame) + ', "output": "' + output_path + '"}')
+
+    except Exception as e:
+        error_msg = str(e).replace('"', '\\"')
```
diane-hj: Could you let me know what is this replacing for? :eyes:

src/keyshot_command_handler.py
```diff
@@ -0,0 +1,158 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+This script runs inside KeyShot. It accepts JSON commands from STDIN and executes them.
+"""
+
+import json
+import sys
+import traceback
+import lux  # noqa: F401
```
diane-hj, line 12: Quick Question - If we wanted to add new KeyShot-specific functionality in the future (like camera controls, material adjustments, etc), we'd primarily extend this file with new command handlers that call the appropriate `lux` API methods? :eyes:

src/session_manager.py
```diff
@@ -0,0 +1,184 @@
...
+        "--keyshot-command-handler-script", required=True, help="Path to KeyShot command handler"
+    )
+    start_parser.add_argument(
+        "--keyshot-bridge-script", required=True, help="Path to KeyShot bridge"
+    )
+    start_parser.add_argument(
+        "--port", type=int, default=1234, help="Port for socket server (default: 1234)"
+    )
+    start_parser.add_argument(
+        "--startup-timeout", type=int, default=60, help="Timeout for KeyShot startup (default: 60)"
```
diane-hj: In the happy cases, I noticed `launch keyshot` takes around 2-5 seconds, which seems pretty quick. In failure cases, user might wait the full 60 seconds without feedback. Would it make sense to share some intermediate progress checks or reduce the timeout? 👀 :

src/keyshot_bridge.py
```diff
@@ -0,0 +1,175 @@
...
+        "keyshot",
+    ]
+
+    for possible_exe in possible_exes:
+        if possible_exe:
+            found_exe = shutil.which(possible_exe)
+            if found_exe:
+                return found_exe
+
+    raise RuntimeError("Cannot find KeyShot executable. Looked for: " + str(possible_exes))
```
diane-hj: Should we use FileNotFoundError? 👀

src/keyshot_bridge.py
```diff
@@ -0,0 +1,175 @@
...
+    parser.add_argument("--scene-file", required=True, help="Path to KeyShot scene file")
+    parser.add_argument(
+        "--keyshot-command-handler-script", required=True, help="Path to KeyShot command handler"
+    )
+    parser.add_argument("--port", type=int, help="Port for socket server")
+    args = parser.parse_args()
+
+    # Validate inputs
+    if not Path(args.scene_file).exists():
+        raise RuntimeError(f"ERROR: Scene file not found: {args.scene_file}")
```
diane-hj: Nit: Could we use FileNotFoundError for missing files? 👀

src/keyshot_command_handler.py
```diff
@@ -0,0 +1,158 @@
...
+                    raise RuntimeError(
+                        "GPU rendering was requested but no compatible GPU is available on this worker. Please manually set min GPUs to 1 under 'Host requirements' in the KeyShot integrated submitter."
+                    )
+        elif render_device == "CPU" and current_device == "GPU":
+            engine_map = {
+                lux.RENDER_ENGINE_PRODUCT_GPU: lux.RENDER_ENGINE_PRODUCT,
+                lux.RENDER_ENGINE_INTERIOR_GPU: lux.RENDER_ENGINE_INTERIOR,
+            }
+            if current_engine in engine_map:
+                lux.setRenderEngine(engine_map[current_engine])
```
diane-hj: I see that we do error handling on line 47, should this part need error handling as well? 👀

src/keyshot_command_handler.py
```diff
@@ -0,0 +1,158 @@
...
+        print(f"Finished Rendering {output_path}")
+        print('{"status": "success", "frame": ' + str(frame) + ', "output": "' + output_path + '"}')
+
+    except Exception as e:
+        error_msg = str(e).replace('"', '\\"')
+        print('{"status": "error", "frame": ' + str(frame) + ', "error": "' + error_msg + '"}')
+
+
+def handle_stop():
+    sys.exit(0)
```
diane-hj, line 120: Nit: can we add print so that we know its handling stop command?

src/keyshot_command_handler.py
crowecawcaw: Yup, exactly!

src/session_manager.py
crowecawcaw: I thought about this more. The time could be longer if the scene is larger. We could try to detect intermediate events based on KeyShot's logs, but I don't think those would be stable long term. I bumped this time to 5 minutes to leave time for larger scenes. If the KeyShot process exists for some reason, it will still quit early.

src/keyshot_command_handler.py
crowecawcaw: These messages were carried over from the previous adapter and unnecessarily messy. Cleaned it up.

src/keyshot_command_handler.py
crowecawcaw: Fixed by combining the two cases

crowecawcaw:
Manually tested:
- rendering a series of frames in sequence
- overriding and not overriding the CPU/GPU settings
- passing through render settings (e.g. samples per pixel) which KeyShot does not save in the scene file

src/task_manager.py
```diff
@@ -0,0 +1,146 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+KeyShot Task Script
+
+This script handles individual render tasks by sending JSON commands to the KeyShot server
+and monitoring STDOUT for progress, errors, and completion.
+
+Usage:
+    python task.py --frame <num> --port <port> --output-path <path> --output-format <format> --render-device <device>
```
diane-hj: should this be `task_manager.py`?

src/session_manager.py
```diff
@@ -0,0 +1,204 @@
...
+        scene_file,
+        "--keyshot-command-handler-script",
+        script_executor,
+    ]
+
+    print("Starting KeyShot bridge...")
+
+    # Find available port and start bridge atomically
+    bridge_process, port = find_available_port_and_start_bridge(bridge_cmd)
+    print(f"KeyShot bridge started. Port:{port} PID:  {bridge_process.pid}")
```
diane-hj: nit. can erase the space front of {bridge_process.pid} :)

src/session_manager.py
```diff
@@ -0,0 +1,204 @@
...
+    print(f"Waiting for KeyShot server to be ready (timeout: {STARTUP_TIMEOUT_SECONDS}s)...")
+    await connect_and_validate(port)
+
+    # Check if bridge process is still running
+    if bridge_process.poll() is not None:
+        raise RuntimeError(
+            f"WARNING: Bridge process (PID: {bridge_process.pid}) has exited with code: {bridge_process.returncode}"
+        )
+
+    print("Session manager completed successfully")
```
diane-hj: nit. can we update the log to say `Session manager started successfully` or `Session manager setup completed successfully`? It could only be me but `completed` could be ambiguous - it's unclear whether this means the session manager finished initializing or the entire session has ended.

diane-hj (dismissed):
LGTM :) I just left a few Nit comments. Thank you!

.github/workflows/release_publish.yml
```diff
@@ -67,17 +67,8 @@ jobs:
...
-    needs: Publish
-    runs-on: ubuntu-latest
-    environment: release
-    name: “Is the Conda Package available in all ProdWaves and have you ran any required manual tests?”
-    steps:
-      - run: |
-         :
-
   ReleaseInstaller:
-    needs: [TagRelease, IsCondaReady]
```
baxeaz, line 80: This is all intended to be getting removed?  I know we're getting rid of the adaptor package - that's basically all this was dealing with previously I guess?

src/keyshot_bridge.py
```diff
@@ -0,0 +1,193 @@
...
+        if possible_exe:
+            found_exe = shutil.which(possible_exe)
+            if found_exe:
+                return found_exe
+
+    raise FileNotFoundError("Cannot find KeyShot executable. Looked for: " + str(possible_exes))
+
+
+async def start_socket_server(process: Process, port: int) -> tuple[Server, asyncio.Task]:
+    # Create shared output queue for all clients
```
baxeaz: Is it actually expected that there'd ever be more than one client?

src/session_manager.py
```diff
@@ -0,0 +1,204 @@
...
+            await asyncio.sleep(1)
+            continue
+
+    raise RuntimeError("Timeout: Could not connect to bridge or receive ready message")
+
+
+async def send_stop_command() -> None:
+    port = read_session_info()
+    try:
+        _reader, writer = await asyncio.open_connection("localhost", port)
```
baxeaz, line 116: Ah interesting, it does indeed look like multiple clients are expected, I guess this is like the environment cleanup action or something?

src/task_manager.py
```diff
@@ -0,0 +1,146 @@
...
+import asyncio
+import json
+import re
+import sys
+import time
+from pathlib import Path
+
+
+SESSION_INFO_FILE = "session_info.json"
+TASK_TIMEOUT_SECONDS = 24 * 60 * 60  # 24 hours
```
baxeaz, line 25: Cancelled/failed jobs should be handled gracefully regardless of what this is set to correct?  Job ends - the stop command should be executed as part of an onExit action for an Environment of some kind maybe?

test/integ/cube/expected_bundle/template.json
```diff
@@ -1,191 +1,217 @@
...
+                                "cancelation": {
+                                    "mode": "NOTIFY_THEN_TERMINATE"
+                                },
+                                "timeout": 390
+                            },
+                            "onExit": {
+                                "command": "python",
+                                "args": [
+                                    "{{Env.File.sessionManagerScript}}",
+                                    "stop"
```
baxeaz, line 157: Ah yeah looks like step environment enter/exit is indeed how this is all stitched together

src/session_manager.py
```diff
@@ -0,0 +1,204 @@
...
+import asyncio
+import json
+import os
+import subprocess
+import sys
+import time
+from pathlib import Path
+
+
+SESSION_INFO_FILE = "session_info.json"
```
baxeaz: Somewhat generic name here - wonder if it might be worth making a bit more keyshot specific to avoid future collisions

src/README.md
```diff
@@ -0,0 +1,214 @@
...
+    TM[Task Manager]
+    KB[KeyShot Bridge]
+    KS["KeyShot Process<br/>(running command handler)"]
+
+    SM -.->|TCP Socket| KB
+    TM -->|TCP Socket| KB
+    KB -->|STDIN/STDOUT| KS    
+```
+
+### Component Specifications
```
baxeaz, line 79: It took me reading through the code (And seeing a test) to understand that the bridge and keyshot was still intended to be launched in a step onEnter and stopped in the onExit - I think that'd be worth calling out in this doc.

.github/workflows/release_publish.yml
crowecawcaw: The PyPi publishing is being removed. On this line, I'm removing `IsCondaReady` which is a manual approval step meant to remind devs to make sure the everything is live in the service before we publish an update. No longer necessary to check though because the adapter Conda package is being dropped.

src/keyshot_bridge.py
crowecawcaw: There should only be one client at a time, but there will one client for each task. The task client connects, runs a command, then disconnects. The queue is necessary so that KeyShot's STDOUT buffer doesn't fill up when there's not a client reading it. I saw KeyShot crashes without it. Added a comment to explain it.

crowecawcaw: Also added a comment explaining who the "clients" are.

src/session_manager.py
crowecawcaw: Correct. I thought about breaking this "session manager" script into "start session" and "end session" scripts so each script did a single thing, but it seemed like more scripts might make it worse.

src/task_manager.py
crowecawcaw: Correct. I did a quick manual check and it did cancel correctly then shutdown KeyShot.

src/submitter.py
```diff
@@ -115,6 +115,18 @@ def construct_job_template(filename: str) -> dict:
         "specificationVersion": "jobtemplate-2023-09",
         "name": filename,
         "parameterDefinitions": [
+            {
+                "name": "AdaptorScriptsDir",
+                "description": "Directory containing adaptor scripts.",
+                "userInterface": {
+                    "control": "HIDDEN"
```
diane-hj: Just curious. Could you let me know what this line do? `control: HIDDEN`

src/submitter.py
```diff
@@ -115,6 +115,18 @@ def construct_job_template(filename: str) -> dict:
...
+                },
+                "type": "PATH",
+                "objectType": "DIRECTORY",
+                "dataFlow": "IN",
+                "default": "adaptor",
```
diane-hj, line 129: I like this approach! Just one clarification question, instead of copying over, we are copying over the script to `adaptor` directory and submitting the directory as part of job right?

diane-hj (dismissed):

src/submitter.py
crowecawcaw: This tells the submitter GUI not to show an input field for this directory, so end users don't wonder what it is or overwrite it.

src/submitter.py
crowecawcaw: Yes, exactly!

src/README.md
```diff
@@ -0,0 +1,214 @@
...
+```json
+{
+  "command": "stop"
+}
+```
+
+
+## Useful testing commands
+
+1. **Build and install the submiter**. Submitter is installed and ready to run in KeyShot. No restart needed.
```
diane-hj: nit.Submitter :)

diane-hj (dismissed):

diane-hj (approved):

baxeaz (approved):

PR #245, opened 2025-09-22
chore: pin the version of click that hatch uses to <8.3 due to Sentin…

moorec-aws:
related: aws-deadline/.github#56

----


crowecawcaw (approved):
Would be better with a code comment explaining why this restriction is there.

epmog (approved):

PR #256, opened 2025-10-14
fix: update submitter path for installer, add instructions

crowecawcaw:
The installer script needs to be updated to use new submitter script path.

Update it. Also add instructions for building the installer.

Fixes the installer configuration.

I built the installer, ran it, and verified it saved the submitter in the correct place.

Yes

No
----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -108,5 +108,14 @@ To run the integration tests:
 
 The user guide is generated from the markdown files in `docs/user_guide` and published to GitHub pages. To view the renderd user guide locally, run `hatch run docs:serve` which will open the user guide in your browser.
 
+## Testing the installer
+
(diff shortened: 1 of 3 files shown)


DEVELOPMENT.md
```diff
@@ -108,5 +108,14 @@ To run the integration tests:
 
 The user guide is generated from the markdown files in `docs/user_guide` and published to GitHub pages. To view the renderd user guide locally, run `hatch run docs:serve` which will open the user guide in your browser.
 
+## Testing the installer
+
+1. Download the install builder tool from https://installbuilder.com/ (Evaluation)
```
diane-hj, line 113: Nit. What do you mean by (Evaluation) here? 👀

diane-hj (approved):

karthikbekalp (approved):

PR #266, opened 2025-11-10
doc: add top level links to README

crowecawcaw:
We want to add a link to the user guide from the README.

Add it. Also add links to the service docs and GitHub org page which might also be useful depending on what the user is looking for. Also dropping the license badge because the license is noted elsewhere and it's not important enough to be the first thing on the README.

Easier to find docs.

https://github.com/crowecawcaw/deadline-cloud-for-keyshot/?tab=readme-ov-file#user-guide--service-documentation--deadline-cloud-on-github

Yes

No

----


edwards-aws (approved):
I like it! Looks great!

mwiebe (approved):

leon-li-inspire (approved):

